PR #8, opened 2018-06-06
update for june release in helloworld

zhouwangzw:
## Purpose

* ...

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

### DIFF
--- a/Windows/quickstart-csharp/Program.cs
+++ b/Windows/quickstart-csharp/Program.cs
@@ -30,7 +30,7 @@ namespace CsharpHelloWorld
                 }
                 else
                 {
-                    Console.WriteLine($"We recognized: {result.RecognizedText}");
+                    Console.WriteLine($"We recognized: {result.Text}");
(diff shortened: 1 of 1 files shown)


Windows/quickstart-csharp/Program.cs
```diff
@@ -30,7 +30,7 @@ class Program
                 }
                 else
                 {
-                    Console.WriteLine($"We recognized: {result.RecognizedText}");
```
wolfma61: >RecognizedText [](start = 63, length = 14)

aren't there other sample accessing RecognizedText and need to get updated??

zhouwangzw: All samples, except this quickstrart_csharp and quickstrat_cpp, will be merged from master, which are from samples.zip. Only these two orphan files are not tested with build. We need to include them into some build process.

PR #17, opened 2018-06-27
Windows/quickstart-{cpp,csharp}: add sample service region, like else…

mahilleb-msft:
Quickstart: give an example for a valid region.

### DIFF
--- a/Windows/quickstart-cpp/CppHelloSpeech.cpp
+++ b/Windows/quickstart-cpp/CppHelloSpeech.cpp
@@ -11,7 +11,7 @@ using namespace Microsoft::CognitiveServices::Speech;
 
 void recognizeSpeech() {
     wstring subscriptionKey{ L"<Please replace with your subscription key>" };
-    wstring region{ L"<Please replace with your service region>" };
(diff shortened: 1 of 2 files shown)


Windows/quickstart-cpp/CppHelloSpeech.cpp
```diff
@@ -11,7 +11,7 @@ using namespace Microsoft::CognitiveServices::Speech;
 
 void recognizeSpeech() {
     wstring subscriptionKey{ L"<Please replace with your subscription key>" };
-    wstring region{ L"<Please replace with your service region>" };
+    wstring region{ L"<Please replace with your service region (e.g., 'westus')>" };
```
zhouwangzw: We need similar changes in our main repo, right?

mahilleb-msft: Yes.

mahilleb-msft: (these ones aren't in the main repo yet, but soon will be)

PR #118, opened 2018-12-10
Allowing SDK to use token vs subscription key

alexjs:
Currently the page itself doesn't work without entering a subscription key here - even if you have a properly configured token.php file.

I've removed these lines because that restores what I believe is the expected functionality, but I realise this leaves a hanging 'Enter your subscription key here' (which is not needed).

It may be better to reject my PR and add in better handling of 'if you have token.php / auth token then use that, ELSE you can enter a subscription key here'.

## Purpose
Changes from requiring subscription key being entered through the web page to properly using the token.php's provided Authorization Token

## Pull Request Type
Removal of (presumably) unused globally scoped vars

```
[x ] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

### DIFF
--- a/samples/js/browser/index.html
+++ b/samples/js/browser/index.html
@@ -200,10 +200,7 @@
                 a.send("");
                 a.onload = function() {
                     var token = JSON.parse(atob(this.responseText.split(".")[1]));
-                    regionKey.value = token.region;
                     authorizationToken = this.responseText;
-                    subscriptionKey.disabled = true;
-                    subscriptionKey.value = "using authorization token (hit F5 to refresh)";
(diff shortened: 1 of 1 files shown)


wolfma61:
we will take a look - sorry for the delay ....

fmegen:
the names of the variables have changes and that was not reflected in the token code. We have updated the sample to work with a proper token.php again (will be public shortly).

you can replace the names from "subscriptionKey" to "key". the name for the region needs to match the regionoption.value that is already in the code. that will allow you to use a token source.

wolfma61:
coming very very soon ....

PR #153, opened 2019-02-05
Tune quickstart content

mahilleb-msft:


quickstart/speech-translation/cpp-windows/helloworld.sln
```diff
@@ -1,4 +1,4 @@
-﻿
+
 Microsoft Visual Studio Solution File, Format Version 12.00
```
wolfma61: what got changed here? just the whitespace?

mahilleb-msft: there was a stray BOM

PR #203, opened 2019-04-08
Updates to samples

PanosPeriorellis:
- added all regions to WPF sample
- added an object to parse the JSON on batch sample
- added properties to batch sample to batch URI parameters

## Purpose

* ...

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
[X] Code style update (formatting, local variables)
[X] Documentation content changes
[ ] Other... Please describe:
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -299,5 +299,3 @@ quickstart/csharp-unity/[Ll]ogs/
 # Linux executables
 quickstart/cpp-linux/helloworld
 samples/cpp/windows/console/samples/sample
-*.dll
-*.exe
--- a/.gitignore
+++ b/.gitignore
@@ -301,5 +301,3 @@ quickstart/cpp-linux/helloworld
 samples/cpp/windows/console/samples/sample
 *.dll
 *.exe
-samples/csharp/dotnet-windows/speechtotext-wpf/speechtotext-wpf/objd/amd64/speechtotext-wpf.exe.config
-samples/csharp/dotnet-windows/speechtotext-wpf/speechtotext-wpf/objd/amd64/Microsoft.CognitiveServices.Speech.csharp.xml
--- a/.gitignore
+++ b/.gitignore
@@ -19,6 +19,7 @@
 [Rr]eleases/
 x64/
 x86/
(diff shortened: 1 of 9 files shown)


.gitignore
```diff
@@ -298,3 +298,9 @@ quickstart/csharp-unity/[Ll]ogs/
 # Linux executables
 quickstart/cpp-linux/helloworld
 samples/cpp/windows/console/samples/sample
+*.dll
+*.exe
+samples/csharp/dotnet-windows/speechtotext-wpf/speechtotext-wpf/objd/amd64/speechtotext-wpf.exe.config
```
wolfma61: >otext-wpf/objd/ [](start = 54, length = 15)

on top of the gitignrore file is a section to ignore build results, adding the objd there might be sufficient

.gitignore
```diff
@@ -298,3 +298,9 @@ quickstart/csharp-unity/[Ll]ogs/
...
+samples/csharp/dotnet-windows/speechtotext-wpf/speechtotext-wpf/objd/amd64/whatstheweatherlike.wav
+samples/batch/csharp/App.config
```
wolfma61: that file is part of the project? why do you want to exclude it?

.gitignore
```diff
@@ -298,3 +298,9 @@ quickstart/csharp-unity/[Ll]ogs/
 # Linux executables
 quickstart/cpp-linux/helloworld
 samples/cpp/windows/console/samples/sample
+*.dll
+*.exe
```
wolfma61: these are probably in the objd directory? so they should be excluded anyhow

PanosPeriorellis: you mean they would be excluded even without me explicitly saying so?

.gitignore
PanosPeriorellis: no, accident. well spotted

.gitignore
PanosPeriorellis: fixed gitignore as per wolfgang's instructions

PanosPeriorellis:
#sign-off

PR #209, opened 2019-04-15
Panosper

PanosPeriorellis:
## Purpose

I added better explanation inside the Readme.md file

## Pull Request Type

it introduces only doc change
```
[ ] Bugfix
[ ] Feature
[x ] Documentation content changes
[ ] Other... Please describe:
```


samples/batch/python/README.md
```diff
@@ -1 +1,15 @@
 Examples in C# and Python on how to use the Speech Services Batch Transcription API.
+
+To execute the sample you need to generate the Python client library.
+
+To do this following the steps:
```
mahilleb-msft:
```suggestion
To do this follow these steps:
```

samples/batch/python/README.md
```diff
@@ -1 +1,15 @@
...
+
+1. Go to https://editor.swagger.io
+2. Click File, then click Import
+3. Enter the swagger URL including the region for your Speech Services subscription https://<your-region>.cris.ai/docs/v2.0/swagger
```
mahilleb-msft:
```suggestion
3. Enter the Swagger URL including the region for your Speech Services subscription: https://<your-region>.cris.ai/docs/v2.0/swagger
```

samples/batch/python/README.md
```diff
@@ -1 +1,15 @@
 Examples in C# and Python on how to use the Speech Services Batch Transcription API.
+
+To execute the sample you need to generate the Python client library.
+
+To do this following the steps:
+
+1. Go to https://editor.swagger.io
+2. Click File, then click Import
```
mahilleb-msft:
```suggestion
2. Click **File**, then click **Import URL**
```

samples/batch/python/README.md
```diff
@@ -1 +1,15 @@
...
+3. Enter the swagger URL including the region for your Speech Services subscription https://<your-region>.cris.ai/docs/v2.0/swagger
+4. Click Generate Client and select Python
```
mahilleb-msft:
```suggestion
4. Click **Generate Client** and select **Python**
```

samples/batch/python/README.md
```diff
@@ -1 +1,15 @@
...
+5. Save the client library
+
+Copy the library in the folder python-library.
+
+You can use a development environment like PyCharm to edit, debug, execute the sample.
```
mahilleb-msft:
```suggestion
You can use a development environment like PyCharm to edit, debug, and execute the sample.
```

samples/batch/python/README.md
```diff
@@ -1 +1,15 @@
 Examples in C# and Python on how to use the Speech Services Batch Transcription API.
```
mahilleb-msft: Why is C# mentioned here?

mahilleb-msft: (And maybe "Example" instead of "Examples"?

samples/batch/python/README.md
```diff
@@ -1 +1,15 @@
...
+
+To do this following the steps:
+
+1. Go to https://editor.swagger.io
+2. Click File, then click Import
+3. Enter the swagger URL including the region for your Speech Services subscription https://<your-region>.cris.ai/docs/v2.0/swagger
+4. Click Generate Client and select Python
+5. Save the client library
+
+Copy the library in the folder python-library.
```
mahilleb-msft: Extract as well?
Should it be in its own directory?
If not, what about the files that are in the generated .zips and the files that are already in here?
For examples, requirements.txt is in the generated .zip, but there's one in the samples directory already - should it be replaced? or not?

samples/batch/python/README.md
PanosPeriorellis: correct, changing it.

samples/batch/python/README.md
PanosPeriorellis: no it should not be on its own directory. the files there remain intact. this is about copying the entire folder (client library) into the python sample project you download.

samples/batch/python/README.md
```diff
@@ -1 +1,15 @@
-Examples in C# and Python on how to use the Speech Services Batch Transcription API.
+Example in Python on how to use the Speech Services Batch Transcription API.
+
+To execute the sample you need to generate the Python client library.
+
+To do this follow these steps:
+
+1. Go to https://editor.swagger.io
+2. Click **File**, then click **Import URL**
+3. Enter the Swagger URL including the region for your Speech Services subscription: https://<your-region>.cris.ai/docs/v2.0/swagger
```
chlandsi: The `<your-region>` isn't show in the markdown.

samples/batch/python/README.md
chlandsi: I think it would be better to pip-install the python client:

6. Extract the downloaded `python-client-generated.zip` somewhere in your file system.
7. Install the extracted `python-client` module in your python environment using pip: `pip install path/to/package/python-client`.
8. The installed package has the name `swagger_client`. You can check that the installation worked using the command `python -c "swagger_client".

chlandsi: With these changes, the files requirements.txt and setup.py can be deleted. main.py should be on the same level as README.md, or in a separate directory that isn't called python-client, to not create confusion with the downloaded package.

chlandsi:
I think it would be cleaner to pip-install the autogenerated python module:

6. Extract the downloaded `python-client-generated.zip` somewhere on your system.
7. Use pip to install the extracted `python-client` package to your python environment: `pip install path/to/package/python-client`
8. The module is now available under the name `swagger_client`. You can check that the installation worked with the command `python -c "import swagger_client"`, which should result in no error.

chlandsi:
With these changes, the files setup.py and requirements.txt can be deleted. The main.py should either be on the same level as the README.md or in a directory below, which then shouldn't be called `python-client`, not to create confusion with the downloaded package.

samples/batch/python/README.md
PanosPeriorellis: addressed

PR #371, opened 2019-08-30
Batch samples: allow to configure region

chlandsi:
- output transcription id in batch sample
- allow to configure region

### DIFF
--- a/samples/batch/csharp/program.cs
+++ b/samples/batch/csharp/program.cs
@@ -77,7 +77,7 @@ namespace BatchClient
 
             Console.WriteLine("Checking status.");
 
-            // check for the status of our transcriptions every 5 sec. (can also be 1, 2, 5 min depending on usage)
+            // check for the status of our transcriptions periodically
             int completed = 0, running = 0, notStarted = 0;
             while (completed < 1)
             {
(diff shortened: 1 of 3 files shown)


samples/batch/csharp/program.cs
```diff
@@ -66,16 +66,18 @@ static async Task TranscribeAsync()
...
             var createdTranscriptions = new List<Guid>();
-            createdTranscriptions.Add(new Guid(transcriptionLocation.ToString().Split('/').LastOrDefault()));
+            createdTranscriptions.Add(thisTranscriptionGuid);
+
+            Console.WriteLine($"Created transcription with id {thisTranscriptionGuid}.");
 
             Console.WriteLine("Checking status.");
 
-            // check for the status of our transcriptions every 30 sec. (can also be 1, 2, 5 min depending on usage)
+            // check for the status of our transcriptions every 5 sec. (can also be 1, 2, 5 min depending on usage)
```
mahilleb-msft: Can you say here "// Check for the status of our transcriptions periodically" and put the timing comments where the actual wait is? "// Check very 5 seconds (adjust according to your usage)" or smth

chlandsi: much better, thanks!

PR #381, opened 2019-09-22
Error handling for the cURL HTTP transfer

famzah:
Indicating the cURL errors should be better than simply getting a blank output from "token.php".


glharper:
This file no longer exists in our repo, and we've switched to a node server pattern to generate auth tokens. Closing.

PR #529, opened 2020-03-04
public samples update for 1.10.0 release.

BrianMouncer:
## Purpose
this is the public samples update for the 1.10.0 release.

## Pull Request Type
new public samples.

```
[ ] Bugfix
[ ] Feature
[ x] Other... Please describe:
```


samples/java/jre/console/src/com/microsoft/cognitiveservices/speech/samples/console/Main.java
```diff
@@ -122,6 +123,8 @@ public static void main(String[] args) {
                 case "p":
                     SpeechSynthesisSamples.synthesisWordBoundaryEventAsync();
                     break;
+                case "q":
+                    SpeechSynthesisScenarioSamples.synthesisServerScenarioAsync();
```
hakulinenjarno: break missing here

BrianMouncer: I checked, and this is the same as in the internal master branch, so it's not a bad merge.
We should file a bug, fix it, and include it in the next update.

If you want, we can fix it directly here in the public repot, so it is fixed before the next release.

---
In reply to: [387453222](https://github.com/Azure-Samples/cognitive-services-speech-sdk/pull/529#discussion_r387453222) [](ancestors = 387453222)

jhakulin: Yes, please fix it in next update

PR #820, opened 2020-09-30
Update transcription enabled storage project

HenryvanderVegte:
## Purpose
Updating transcription enabled storage project - refactoring and improved retry logic, improved scalability

## Pull Request Type
```
[ ] Bugfix
[ ] Feature
[X] Code style update (formatting, local variables)
[X] Refactoring (no functional changes, no api changes)
[ ] Other... Please describe:
```


samples/batch/transcription-enabled-storage/Connector/BatchClient.cs
```diff
@@ -7,67 +7,67 @@ namespace Connector
...
-            var path = $"{HostName}{SpeechToTextBasePath}Transcriptions/{id}/files";
+            var path = $"{transcriptionLocation}/files";
 
             var combinedTranscriptionFiles = new List<TranscriptionFile>();
 
             do
             {
-                var transcriptionFiles = await this.GetAsync<TranscriptionFiles>(path).ConfigureAwait(false);
+                // Delay to avoid throttling
+                await Task.Delay(2000).ConfigureAwait(false);
```
zhouwangzw, line 68: do you need exponential backoff, or at least  increase the delay time based on retry times.

zhouwangzw: ok, this is just wait for retrieving the nextlink. Is this really needed? I would avoid to use hardcoded delay here, since it just slow the processing. What we should do is adding retry in http calls in case of Status code is 429 or something else. You can use polly like CRIS backend, or manually do that.

samples/batch/transcription-enabled-storage/Connector/BatchClient.cs
```diff
@@ -7,67 +7,67 @@ namespace Connector
...
+
+        public static Task<SpeechTranscript> GetSpeechTranscriptFromSasAsync(string sasUri, ILogger log)
         {
-            var path = $"{HostName}{SpeechToTextBasePath}Transcriptions";
-            return this.GetAsync<IEnumerable<Transcription>>(path);
+            return GetAsync<SpeechTranscript>(sasUri, null, DefaultTimeout, log);
         }
 
-        public Task<Transcription> GetTranscriptionAsync(Guid id)
+        public static Task<Transcription> GetTranscriptionAsync(string transcriptionLocation, string subscriptionKey, ILogger log)
```
zhouwangzw, line 54: Do we really need to pass subscriptionKey? I would prefer to create a client which is associated with subscritionkey, and so that no need to pass subscriptionkey as parameter

samples/batch/transcription-enabled-storage/Setup/ArmTemplate.json
```diff
@@ -190,7 +177,7 @@
             }
         },
         "PunctuationMode": {
-            "defaultValue": "DictatedAndAutomatic",
+            "defaultValue": "None",
```
zhouwangzw, line 180: Is by intention to set punctuation mode to None? The default mode for transcription is DictatedAndAutomatic.

zhouwangzw (dismissed):

samples/batch/transcription-enabled-storage/Setup/ArmTemplate.json
HenryvanderVegte: okay, I'll set the default to DictatedAndAutomatic. Is the default for ProfanityFIlterMode also != None?

samples/batch/transcription-enabled-storage/Connector/BatchClient.cs
HenryvanderVegte: I just thought that since I'm sending multiple requests one after another here against the service, I'm trying to avoid getting throttled. 
I think I can just remove it, because if I'm doing a GET Files, I'm either getting a fast response if there are only a few files - or a slow response for many files. And if I get a slow response, throttling is probably not a problem anyway.

samples/batch/transcription-enabled-storage/Connector/BatchClient.cs
HenryvanderVegte: I was following the Manage connections guidelines (https://docs.microsoft.com/en-gb/azure/azure-functions/manage-connections) to use static clients for http and storage connections. Of course I could have an instance of a BatchClient that references the static http client, but I thought that this might look a bit more messy.

zhouwangzw: ok, that makes sense for Azure Functions.

zhouwangzw (approved):

PR #843, opened 2020-10-23
Add example how to use Custom Translation models

nkgami:
## Purpose

* https://docs.microsoft.com/en-us/azure/cognitive-services/speech-service/speech-translation
> These samples cover common scenarios like reading audio from a file or stream, continuous and single-shot 
recognition/translation, and working with custom models.

This document mentions about Custom Models, but there seems no sample for Custom Translator. So I added the example.

## Pull Request Type

```
[ ] Bugfix
[x] Feature
[ ] Other... Please describe:
```

### DIFF
--- a/samples/js/browser/index.html
+++ b/samples/js/browser/index.html
@@ -209,6 +209,12 @@
                 </td>
             </tr>
             <tr>
+                <td align="right">Category ID of Custom Translator:</td>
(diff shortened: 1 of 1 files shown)


nkgami:
Moved to #846 because this code does not cover some use scenario

PR #903, opened 2020-12-22
JavaScript example with interim results

TakamotoAI:
Please add to the JavaScript example the very demo you have on your website showing real-time transcription with interim results in web browser


glharper:
Why not update all version numbers to 1.18.0?

Issue #933, opened 2021-01-22
How to save output to file and download


TwinMist:
Hi, I am no coder, and wondering if someone could help me.
using the synthesis.html
https://github.com/Azure-Samples/cognitive-services-speech-sdk/blob/master/samples/js/browser/synthesis.html
How can i save the output that is currently played though the speaker to also a file and also let to user to download the file saved.
i would be very greatful for some help
thanks

glharper:
@TwinMist Thanks for using Speech SDK and writing up this issue. For browsers, there's not an easy way to do direct file writing because of the security separation between the browser and the client file system. I would suggest an approach along these lines, but how you'd like to actually create a file and present it for download is up to you. Each audio sample from the callback is playable, so I can't tell you how to combine them all into one file. You can set the desired audio format using 
speechConfig.speechSynthesisOutputFormat = SpeechSDK.SpeechSynthesisOutputFormat.<DESIRED_FORMAT>;
(See [here](https://github.com/microsoft/cognitive-services-speech-sdk-js/blob/master/src/sdk/SpeechSynthesisOutputFormat.ts) for all available formats)
```
const speechConfig = SpeechSDK.SpeechConfig.fromSubscription(subscriptionKey, serviceRegion);
// For audio voice output
speechConfig.setProperty(SpeechSDK.PropertyId.SpeechServiceConnection_TranslationVoice, "en-US");
const synthesizer = new SpeechSDK.SpeechSynthesizer(speechConfig);
let buffers = new ArrayBuffer[];
synthesizer.synthesizing = (s, e) => {
 let newData = e.result.audioData;
 let audioSize = newData === undefined ? 0 : newData.byteLength;

 if (audioSize > 0) {
 // each newData is a playable audio sample, with appropriate headers (in the case of .wav and .ogg files) added.
 buffers.push(newData);
 }
};
synthesizer.speakTextAsync(
 "<YOUR_TEXT_HERE>",	  
 function (result) {
  window.console.log(result);
  if (result.reason === SpeechSDK.ResultReason.SynthesizingAudioCompleted) {
   // do something here to collect ArrayBuffers in buffers and present to user as downloadable file
  } 
 },
 function (err) {
  window.console.log(err);
 });
```

TwinMist:
OK bit lost, is there a way after the audio is played it then just prompts the user to download and save it?
[...]

PR #938, opened 2021-01-26
add download synthesized audio in TTS JS sample

yulin-li:
## Purpose

* ...

## Pull Request Type

```
[ ] Bugfix
[x] Feature
[ ] Other... Please describe:
```
add download synthesized audio in TTS JS sample, for #933


glharper (approved):

TwinMist:
Is it possible to change the format of the saved file? would like to use raw-8khz-8bit-mono-mulaw so i dont have to convert it from mp3

thanks

warrenkc:
The download is missing the voice name. You need to add this code below to line 445 in the file. Otherwise it uses the default voice not the selected voice for the download.

speechConfig.speechSynthesisVoiceName = voiceOptions.value;

PR #942, opened 2021-02-01
add retry + consolidated output

HenryvanderVegte:
## Purpose
* Add retry for sas urls
* Don't fail function if transcript already exists on storage
* Write job id and region to storage if job failed

## Pull Request Type

```
[x] Bugfix
[x] Feature
[ ] Other... Please describe:
```


chlandsi (approved):

samples/batch/transcription-enabled-storage/Connector/StorageConnector.cs
```diff
@@ -119,17 +129,24 @@ public async Task WriteTextFileToBlobAsync(string content, string containerName,
...
+            var blockBlobClient = container.GetBlobClient(fileName);
+
+            if (await blockBlobClient.ExistsAsync().ConfigureAwait(false))
+            {
+                log.LogError($"File {fileName} already exists in container {containerName}. Returning.");
+                return;
+            }
 
             using (var stream = new MemoryStream(Encoding.UTF8.GetBytes(content)))
             {
```
zhouwangzw, line 141: instead of using  stream, I would prefer to use UpdateTextAsync() to write the file with one-shot.  For streaming, it is possible that the file is only partially written.

samples/batch/transcription-enabled-storage/Connector/StorageConnector.cs
```diff
@@ -142,8 +159,18 @@ public async Task MoveFileAsync(string inputContainerName,  string inputFileName
             var outputContainerClient = BlobServiceClient.GetBlobContainerClient(outputContainerName);
             var outputBlockBlobClient = outputContainerClient.GetBlobClient(outputFileName);
 
+            if (await outputBlockBlobClient.ExistsAsync().ConfigureAwait(false))
```
zhouwangzw, line 162: What happens if the file is only partially written?

samples/batch/transcription-enabled-storage/FetchTranscription/TranscriptionProcessor.cs
```diff
@@ -67,14 +67,14 @@ public static async Task ProcessTranscriptionJobAsync(TranscriptionStartedMessag
             }
             catch (WebException e)
             {
-                if (e.Response != null && BatchClient.IsThrottledOrTimeoutStatusCode(((HttpWebResponse)e.Response).StatusCode))
+                if (e.Response != null && (BatchClient.IsThrottledOrTimeoutStatusCode(((HttpWebResponse)e.Response).StatusCode) || ((HttpWebResponse)e.Response).StatusCode == HttpStatusCode.InternalServerError))
```
zhouwangzw, line 70: Do you need any backoff time if it is Internal Server error?

zhouwangzw (approved):

samples/batch/transcription-enabled-storage/FetchTranscription/TranscriptionProcessor.cs
HenryvanderVegte: we got backoff time already in there since we exponentially retry in that case here

samples/batch/transcription-enabled-storage/Connector/StorageConnector.cs
HenryvanderVegte: right.. probably I should override it here

samples/batch/transcription-enabled-storage/Connector/StorageConnector.cs
HenryvanderVegte: Update/UploadTextAsync was part of microsoft.windowsazure.storage I think, at least I don't have it in azure.storage.blobs anymore

zhouwangzw (approved):

Issue #960, opened 2021-02-12
Option in js sample to select output


TwinMist:
Be great if within the js sample there could be a option to select the type of output format when saving the audio

pankopon:
Hi, so this must be about https://github.com/Azure-Samples/cognitive-services-speech-sdk/blob/master/samples/js/browser/synthesis.html - I have added a work item about it to our backlog for consideration, and we will update this item when the change is available.

pankopon:
Closing as added to the backlog.

PR #972, opened 2021-03-02
Add option in JS sample to select output format

yulin-li:
## Purpose
Address #960 , add option in JS sample to select output format
```
[ ] Bugfix
[x] Feature
[ ] Other... Please describe:
```

image


TwinMist:
When you change the format and you try and save the file extension it is always .mp3 so if you select Raw8Khz8BitMonoMULaw it should be .wav.

PR #1035, opened 2021-04-07
Minor updates to UWP samples

dargilco:
Add full list of supported regions in the UWP sample combo box.
Use "Microsoft" as the publisher display name in UWP sample & quickstart.


BrianMouncer (approved):
:shipit:

BrianMouncer (approved):

PR #1274, opened 2021-10-05
Update document to enable differentiate guest speakers

jimxieJ:
## Purpose
Since now the CTS service sets differentiate guest speakers as disable by default and users have to enable it explicitly, I added the setup to enable differentiate guest speakers in sample code.

## Pull Request Type

```
[ ] Bugfix
[x] Feature
[ ] Other... Please describe:
```

### DIFF
--- a/quickstart/csharp/dotnet/conversation-transcription/helloworld/Program.cs
+++ b/quickstart/csharp/dotnet/conversation-transcription/helloworld/Program.cs
@@ -79,6 +79,10 @@ namespace helloworld
         {
             var config = SpeechConfig.FromSubscription(subscriptionKey, region);
             config.SetProperty("ConversationTranscriptionInRoomAndOnline", "true");
+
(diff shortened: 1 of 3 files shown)


glharper (approved):

glharper (changes requested):
Changing approval to a request to wait and discuss.

BrianMouncer (approved):
This looks ok, but does it drastically change the sample that was there before?
should we instead make this an option that can be turned on and off via some menu or UI widget?  That was the user can experiment with and without the feature turned on...

v-demjoh (changes requested):
don't add string constant keys like "DifferentiateGuestSpeakers" to public-facing information. there's an internal wiki where we list known publications of string constant keys, but we want to shorten that list.

jimxieJ:
Let's wait for Speech SDK Team Decision.

PR #1279, opened 2021-10-11
[Ingestion Client] Add package versioning + byos SAS

HenryvanderVegte:
## Purpose
Add versioning for app packages + parameter whether SAS should be generated as part of the request

## Pull Request Type

```
[ ] Bugfix
[x] Feature
[ ] Other... Please describe:
```


PanosPeriorellis (approved):
i saw the logic. we assume that when a BYOS is set up is also correct. Do you think we should have a fall back? in case the absolute URI does not work?

PR #1301, opened 2021-11-09
Add snippet tag

eric-urban:
Snippet will be referenced from [Use a keyword model with the Speech SDK](https://docs.microsoft.com/en-us/azure/cognitive-services/speech-service/custom-keyword-basics?pivots=programming-language-python#use-a-keyword-model-with-the-speech-sdk).


samples/python/console/speech_sample.py
```diff
@@ -288,6 +288,7 @@ def stop_cb(evt):
     speech_recognizer.stop_continuous_recognition()
     # </SpeechContinuousRecognitionWithFile>
 
+# <SpeechRecognitionFromMicrophone>
```
rhurey, line 291: Seems this tag should have keyword in it since it's the keyword sample.

eric-urban: @rhurey - Updated. Thanks

eric-urban:
@rhurey @dargilco - can you please merge?

PR #1326, opened 2021-11-22
Samples for detailed reco results, including word-level timing

dargilco:
Samples for detailed reco results, including word-level timing (C++, C#, Java, Python)

This is a copy of the changes already made and reviewed in our internal repo.

### DIFF
--- a/samples/python/console/speech_sample.py
+++ b/samples/python/console/speech_sample.py
@@ -10,6 +10,7 @@ Speech recognition samples for the Microsoft Cognitive Services Speech SDK
 import time
 import wave
 import string
+import json
 
 try:
     import azure.cognitiveservices.speech as speechsdk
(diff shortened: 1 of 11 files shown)


samples/python/console/speech_sample.py
```diff
@@ -92,6 +93,69 @@ def speech_recognize_once_from_file():
...
+
+        # Time units are in hundreds of nanoseconds (HNS), where 10000 HNS equals 1 millisecond
+        print("Offset: {}".format(result.offset))
+        print("Duration: {}".format(result.duration))
+
+        # Now get the detailed recognition results from the JSON
+        json_result = json.loads(result.json)
+
+        # The first cell in the NBest list corresponds to the recognition results 
+        # (NOT the cell with the highest confidence number!)
```
glharper, line 136: Can we put a link to the FAQ in the comments here?

glharper (approved):

samples/python/console/speech_sample.py
dargilco: @glharper Good idea! I'm going to write a new public doc on detailed reco result, so I will make a separate change in this sample in the Carbon repo to link to the new page, or link to the existing FAQ as you suggested if the new page is taking too long to write.

PR #1336, opened 2021-11-29
updated with Intent recognition pattern matching sample

chschrae:
## Purpose

Added sample for the Intent Recognizer changes released in 1.19

## Pull Request Type

```
[ ] Bugfix
[x] Feature
[ ] Other... Please describe:
```

### DIFF
--- a/samples/cpp/windows/console/samples/intent_recognition_samples.cpp
+++ b/samples/cpp/windows/console/samples/intent_recognition_samples.cpp
@@ -222,3 +222,86 @@ void IntentContinuousRecognitionWithFile()
     recognizer->StopContinuousRecognitionAsync().get();
     // </IntentContinuousRecognitionWithFile>
 }
+
+// Intent recognition using pattern matching and microphone.
+void IntentRecognitionWithPatternMatchingAndMicrophone()
+{
+    // <IntentRecognitionWithPatternMatchingAndMicrophone>
+    // Creates an instance of a speech config with specified subscription key
+    // and service region. Note that in contrast to the other samples this DOES NOT require a
+    // LUIS application.
+    // The default recognition language is "en-us".
+    auto config = SpeechConfig::FromSubscription("YourLanguageUnderstandingSubscriptionKey", "YourLanguageUnderstandingServiceRegion");
+
+    // Creates an intent recognizer using microphone as audio input.
+    auto recognizer = IntentRecognizer::FromConfig(config);
+
+    // Creates a Pattern Matching model and adds specific intents from your model. 
+    // The Id is used to identify this model from others in the collection.
+    auto model = PatternMatchingModel::FromModelId("YourPatternMatchingModelId");
+
+    // Adds some intents to look for specific patterns.
+    model->Intents.push_back({ {"Take me to floor {floorName}", "Go to floor {floorName}", "Take me to {floorName}", "Go to floor {floorName}", "Go to {floorName}", "{floorName}"}, "ChangeFloors"});
+    model->Intents.push_back({ {"{action} the doors", "{action} doors", "{action} the door", "{action} door"}, "DoorControl"});
+
+    // Creates the "floorName" entity and set it to type list.
+    // Adds acceptable values. NOTE the default entity type is Any and so we do not need
+    // to declare the "action" entity.
+    model->Entities.push_back({ "floorName" , Intent::EntityType::List, Intent::EntityMatchMode::Strict, {"ground floor", "lobby", "1st", "first", "one", "2nd", "second", "two"}});
+
+    // Add the model to a model vector.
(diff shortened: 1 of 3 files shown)


samples/cpp/windows/console/samples/intent_recognition_samples.cpp
```diff
@@ -222,3 +222,86 @@ void IntentContinuousRecognitionWithFile()
     recognizer->StopContinuousRecognitionAsync().get();
     // </IntentContinuousRecognitionWithFile>
 }
+
+// Intent recognition using pattern matching and microphone.
+void IntentRecognitionWithPatternMatchingAndMicrophone()
+{
+    // <IntentRecognitionWithPatternMatchingAndMicrophone>
```
BrianMouncer, line 229: >IntentRecognitionWithPatternMatchingAndMicrophone [](http://example.com/codeflow?start=8&length=49)

Is this a new code tag?
If so you need to update the white list...

BrianMouncer (approved):
:shipit:

samples/cpp/windows/console/samples/intent_recognition_samples.cpp
BrianMouncer: you will notice when you build this in carbon repo.

dargilco (approved):

samples/cpp/windows/console/samples/intent_recognition_samples.cpp
chschrae: will do

samples/cpp/windows/console/samples/intent_recognition_samples.cpp
```diff
@@ -222,3 +222,86 @@ void IntentContinuousRecognitionWithFile()
...
+
+    // Creates an intent recognizer using microphone as audio input.
+    auto recognizer = IntentRecognizer::FromConfig(config);
+
+    // Creates a Pattern Matching model and adds specific intents from your model. 
+    // The Id is used to identify this model from others in the collection.
+    auto model = PatternMatchingModel::FromModelId("YourPatternMatchingModelId");
+
+    // Adds some intents to look for specific patterns.
+    model->Intents.push_back({ {"Take me to floor {floorName}", "Go to floor {floorName}", "Take me to {floorName}", "Go to floor {floorName}", "Go to {floorName}", "{floorName}"}, "ChangeFloors"});
```
ralph-msft, line 244: Nitpick: Would it make sense to move the ID for the intent to be the first item? I'm of the opinion that makes the code a little more readable. Probably a breaking change now though

samples/cpp/windows/console/samples/intent_recognition_samples.cpp
```diff
@@ -222,3 +222,86 @@ void IntentContinuousRecognitionWithFile()
...
+    std::vector<std::shared_ptr<LanguageUnderstandingModel>> modelCollection;
+    modelCollection.push_back(model);
+    
+    // Apply the language model collection to the recognizer.
+    recognizer->ApplyLanguageModels(modelCollection);
+
+    cout << "Say something...\n";
+
+    // Starts intent recognition, and returns after a single utterance is recognized. The end of a
+    // single utterance is determined by listening for silence at the end or until a maximum of 15
```
ralph-msft, line 262: Nitpick: Oxford commas:
"silence at the end, or until"

samples/cpp/windows/console/samples/intent_recognition_samples.cpp
```diff
@@ -222,3 +222,86 @@ void IntentContinuousRecognitionWithFile()
...
+    // seconds of audio is processed.  The task returns the recognition text as result.
```
ralph-msft, line 263: Did you mean future here instead of task?

samples/cpp/windows/console/samples/intent_recognition_samples.cpp
```diff
@@ -222,3 +222,86 @@ void IntentContinuousRecognitionWithFile()
...
+    // Checks result.
+    if (result->Reason == ResultReason::RecognizedIntent)
+    {
+        cout << "RECOGNIZED: Text=" << result->Text << std::endl;
+        cout << "  Intent Id: " << result->IntentId << std::endl;
+
+        auto entities = result->GetEntities();
+        if (entities.find("floorName") != entities.end())
+        {
+            std::cout << "  Floor name: = " << entities["floorName"].c_str() << std::endl;
```
ralph-msft, line 278: Nitpick: std::cout is the preferred way per our coding guidelines. However I'm noticing that we are consistent with when we use the std:: prefix. For example we don't use for cout at line 272 but do use it for the std:endl at the end of that line.

This is potentially outside the scope of this PR, so perhaps we should revisit this separately and make the file more consistent.

PR #1404, opened 2022-02-09
Fix build break in C# sample

dargilco:
This was a result of bad merge following the update of samples for the 1.20 release.


BrianMouncer (approved):
Thanks

PR #1435, opened 2022-03-10
Add captioning samples

v-jaswel:
## Purpose

* Add captioning samples. For now, this is a draft PR for review purposes.

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
[x] Other... Please describe: Add captioning samples.
```


scenarios/cpp/windows/captioning/caption.cpp
```diff
@@ -0,0 +1,580 @@
...
+"              -t: Enable TrueText.\n"
+"              -u: Emit partial results instead of finalized results.\n";
+
+/*
+Types
+*/
+
+struct UserConfig
+{
+	const bool prioritizeAccuracyEnabled = false;
```
dargilco, line 69: We don't use TABS in our code, since it's unpredictable how that shows up in different editors. Replace every tab with 4 spaces. Set your editor to insert 4 spaces whenever you hit the tab key.

scenarios/cpp/windows/captioning/caption.cpp
```diff
@@ -0,0 +1,580 @@
...
+
+/*
+Helper functions
+*/
+
+/* Simple command line argument parsing functions. See
+https://stackoverflow.com/a/868894
+The value parameter to std::find must be std::string. Otherwise std::find will compare the pointer values rather than the string contents.
+*/
+std::optional<std::string> GetCmdOption (char** begin, char** end, const std::string& option)
```
dargilco, line 159: We don't put a space between the method/class name and the opening parenthesis ("GetCmdOption (...")

scenarios/cpp/windows/captioning/caption.cpp
```diff
@@ -0,0 +1,580 @@
...
+	{
+		audioConfig = Audio::AudioConfig::FromDefaultMicrophoneInput ();
+	}
+	return audioConfig;
+}
+
+std::shared_ptr<SpeechConfig> SpeechConfigFromUserConfig (UserConfig userConfig)
+{
+	std::shared_ptr<SpeechConfig> speechConfig;
+	if (userConfig.languageIDLanguages.has_value ()) {
```
dargilco, line 400: Be consistent in putting the open and close curly brackets "{", "}" in a new line

scenarios/cpp/windows/captioning/caption.cpp
```diff
@@ -0,0 +1,580 @@
...
+	const float startMinutes_2 = fmod (startMinutes_1, 60.0);
+	const float endMinutes_2 = fmod (endMinutes_1, 60.0);
+	
+	return CaptionTime (startHours, endHours, startMinutes_2, endMinutes_2, startSeconds_2, endSeconds_2);
+}
+
+std::string AddLanguageToResult (std::shared_ptr<SpeechRecognitionResult> result, UserConfig userConfig)
+{
+	std::ostringstream text;
+	if (true == userConfig.languageIDLanguages.has_value ())
```
dargilco, line 227: Can you look at other C++ sample code? I don't think we use the pattern if (true == X). We just do if (X) and if (!X).

scenarios/cpp/windows/captioning/caption.cpp
```diff
@@ -0,0 +1,580 @@
...
+	- Microsoft.CognitiveServices.Speech.core.dll
+	- Microsoft.CognitiveServices.Speech.extension.audio.sys.dll
+*/
+
+/*
+Dependencies
+*/
+
+// exception
+#include <exception>
```
dargilco, line 12: Can you put the comment at the end of the #include, to make it more compact? I think that's what we do elsewhere. Should be:
#include <exception> // exception

scenarios/cpp/windows/captioning/caption.cpp
```diff
@@ -0,0 +1,580 @@
...
+using namespace Microsoft::CognitiveServices::Speech;
+using namespace Microsoft::CognitiveServices::Speech::Speaker;
+
+/*
+Global constants
+*/
+
+// Note: Continuous language identification is supported only with v2 endpoints.
+const std::string v2EndpointStart ("wss://");
+const std::string v2EndpointEnd (".stt.speech.microsoft.com/speech/universal/v2");
```
dargilco, line 42: Usually, we define const in the file scope when they are used in multiple places (functions). Here you only use them in one function V2EndpointFromRegion(). I think you can move these definitions into that function. 

Same for other consts if they are only used in one place.

dargilco (changes requested):

scenarios/cpp/windows/captioning/caption.cpp
```diff
@@ -0,0 +1,535 @@
...
+    {
+        if (CmdOptionExists(argv, argv + argc, "-h"))
+        {
+            std::cout << usage << std::endl;
+        }
+        else
+        {
+            UserConfig userConfig = UserConfigFromArgs(argc, argv);
+            std::shared_ptr<SpeechRecognizer> speechRecognizer = SpeechRecognizerFromUserConfig(userConfig);
+            Initialize(userConfig);
```
eric-urban, line 518: Can this move within or right after `UserConfigFromArgs`? Currently it breaks up new SpeechRecognizer -> RecognizeContinuous flow.

scenarios/csharp/dotnet/captioning/Program.cs
```diff
@@ -0,0 +1,423 @@
...
+            }
+            
+            return speechRecognizer;
+        }
+
+        static SpeechRecognizer SpeechRecognizerFromUserConfig(UserConfig userConfig)
+        {
+            AudioConfig audioConfig = AudioConfigFromUserConfig(userConfig);
+            SpeechConfig speechConfig = SpeechConfigFromUserConfig(userConfig);
+            return SpeechRecognizerFromSpeechConfig(speechConfig, audioConfig, userConfig);
```
eric-urban, line 290: @v-jaswel - "FromSpeechConfig" does not quite capture how a SpeechRecognizer should be constructed. Also not clear why do we need an extra helper method? I suggest replace the current `return` line with:

```csharp
            SpeechRecognizer speechRecognizer;
            if (userConfig.languageIDLanguages is string[] languageIDLanguagesValue)
            {
                AutoDetectSourceLanguageConfig autoDetectSourceLanguageConfig = AutoDetectSourceLanguageConfig.FromLanguages(languageIDLanguagesValue);
                speechRecognizer = new SpeechRecognizer(speechConfig, autoDetectSourceLanguageConfig, audioConfig);
            }
            else
            {
                speechRecognizer = new SpeechRecognizer(speechConfig, audioConfig);
            }

            if (userConfig.phraseList is string phraseListValue)
            {
                PhraseListGrammar grammar = PhraseListGrammar.FromRecognizer(speechRecognizer);
                grammar.AddPhrase(phraseListValue);
            }

            return speechRecognizer;
```

scenarios/csharp/dotnet/captioning/Program.cs
```diff
@@ -0,0 +1,423 @@
...
+            
+            return speechConfig;
+        }
+
+        static SpeechRecognizer SpeechRecognizerFromSpeechConfig(SpeechConfig speechConfig, AudioConfig audioConfig, UserConfig userConfig)
+        {
+            SpeechRecognizer speechRecognizer;
+            if (userConfig.languageIDLanguages is string[] languageIDLanguagesValue)
+            {
+                AutoDetectSourceLanguageConfig detectLanguageConfig = AutoDetectSourceLanguageConfig.FromLanguages(languageIDLanguagesValue);
```
eric-urban, line 269: Matches the naming convention already in our GitHub samples.
```suggestion
                AutoDetectSourceLanguageConfig autoDetectSourceLanguageConfig = AutoDetectSourceLanguageConfig.FromLanguages(languageIDLanguagesValue);
```

scenarios/csharp/dotnet/captioning/Program.cs
```diff
@@ -0,0 +1,423 @@
...
+              -m: Enable profanity filter (mask profanity). -f overrides this.
+         -o file: Output to *file*.
+      -p phrases: Add specified *phrases*.
+                  Example: Constoso;Jessie;Rehaan
+              -q: Suppress console output (except errors).
+       -r number: Set stable partial result threshold to *number*.
+                  Example: 3
+              -s: Emit SRT (default is WebVTT.)
+              -t: Enable TrueText.
+              -u: Emit partial results instead of finalized results.";
```
eric-urban, line 393: @v-jaswel - Why does it have to be a choice? We could demo how to use both in parallel. While recognizing they can get the offset and intermediate duration, and we can write these to the console. But there are no word level details until you get the final recognized result. The final result could be written as caption.

scenarios/csharp/dotnet/captioning/Program.cs
```diff
@@ -0,0 +1,423 @@
...
+                        }
+                        else if (ResultReason.NoMatch == e.Result.Reason)
+                        {
+                            WriteToConsole("NOMATCH: Speech could not be recognized.\n", userConfig);
+                        }
+                    };
+            }
+            else
+            {
+                speechRecognizer.Recognized += (object? sender, SpeechRecognitionEventArgs e) =>
```
eric-urban, line 315: This might or might not be in scope, but if you want detailed recognition results with word-level offset and duration, set the following:
```csharp
// Get Words with Offset and Duration.
speechConfig.RequestWordLevelTimestamps();
```

Then we could show something like this: 
```csharp
speech_recognizer.Recognized += (object sender, SpeechRecognitionEventArgs e) =>
    {
        if (ResultReason.RecognizedSpeech == e.Result.Reason && e.Result.Text.Length > 0)
        {            
            Console.WriteLine($"RECOGNIZED: Text={e.Result.Text}");
            Console.WriteLine(String.Format ("Offset in Ticks: {0}", e.Result.OffsetInTicks));
            Console.WriteLine(String.Format ("Duration in Ticks: {0}", e.Result.Duration.Ticks));
                        
            var detailedResults = speechRecognitionResult.Best();
            if(detailedResults != null && detailedResults.Any())
            {
                // The first item in detailedResults corresponds to the recognized text.
                // This is not necessarily the item with the highest confidence number.
                var bestResults = detailedResults?.ToList()[0];
                Console.WriteLine($"\tConfidence: {bestResults.Confidence}\n\tText: {bestResults.Text}\n\tLexicalForm: {bestResults.LexicalForm}\n\tNormalizedForm: {bestResults.NormalizedForm}\n\tMaskedNormalizedForm: {bestResults.MaskedNormalizedForm}");
                
                // You must set speechConfig.RequestWordLevelTimestamps() to get word-level timestamps.
                Console.WriteLine($"\tWord-level timing:");
                Console.WriteLine($"\t\tWord | Offset | Duration");
                Console.WriteLine($"\t\t----- | ----- | ----- ");

                foreach (var word in bestResults.Words)
                {
                    Console.WriteLine($"\t\t{word.Word} | {word.Offset} | {word.Duration}");
                }
            }
        }
    };
```

scenarios/csharp/dotnet/captioning/Program.cs
v-jaswel: Yeah I just figured they didn't want their caption output file cluttered up with partial results. I could make it so that partial results are always written only to the console, never to file. I don't know how the output is going to be consumed in the final product.

scenarios/csharp/dotnet/captioning/Program.cs
v-jaswel: I don't know whether they want word-level timestamping. I'll leave this open until we know. This sounds like a good thing to bring up in today's meeting or in that e-mail thread if Katerina responds.

scenarios/csharp/dotnet/captioning/Program.cs
```diff
@@ -0,0 +1,417 @@
+/*
```
trrwilson, line 1: Do we need the standard copyright header on these? E.g.: https://github.com/Azure-Samples/cognitive-services-speech-sdk/blob/master/quickstart/csharp/dotnetcore/from-microphone/helloworld/Program.cs

scenarios/csharp/dotnet/captioning/Program.cs
```diff
@@ -0,0 +1,417 @@
...
+Dependencies
```
trrwilson, line 2: Categorizing pieces of code like the `using` statements seems a little too guided; I'd generally suggest we stay away from adding documentation for anything but our product usage and *significantly* non-intuitive non-product stuff--standard `using` boilerplate is table stakes enough that it needs no intro (and if it does, we've got other problems)

scenarios/csharp/dotnet/captioning/Program.cs
```diff
@@ -0,0 +1,417 @@
...
+/* Notes:
+- Install the Speech NuGet package:
+dotnet add package Microsoft.CognitiveServices.Speech
+- The Speech SDK on Windows requires the Microsoft Visual C++ Redistributable for Visual Studio 2019 on the system. See:
+https://docs.microsoft.com/azure/cognitive-services/speech-service/speech-sdk?tabs=windows%2Cubuntu%2Cios-xcode%2Cmac-xcode%2Candroid-studio
+*/
+using Microsoft.CognitiveServices.Speech;
+using Microsoft.CognitiveServices.Speech.Audio;
+
+namespace Caption
```
trrwilson, line 19: It doesn't get more nitpicky than this, but I'd suggest "Captions" or "Captioning" for the namespace -- probably "Captioning" given the folder structure

scenarios/csharp/dotnet/captioning/Program.cs
```diff
@@ -0,0 +1,417 @@
...
+            this.partialResultsEnabled = partialResultsEnabled;
+            this.stablePartialResultThreshold = stablePartialResultThreshold;
+            this.srtEnabled = srtEnabled;
+            this.trueTextEnabled = trueTextEnabled;
+            this.subscriptionKey = subscriptionKey;
+            this.region = region;
+        }
+    }
+
+    public class TooFewArguments : Exception
```
trrwilson, line 73: Derived classes of `Exception` should generally still have "Exception" on the ends of their type names, e.g. all the built-in stuff visualized here: https://www.tutorialsteacher.com/csharp/csharp-exception

trrwilson: For this, it's probably just fine to use `ArgumentException` with a clear message

scenarios/csharp/dotnet/captioning/Program.cs
```diff
@@ -0,0 +1,417 @@
...
+    }
+
+    class Program
+    {
+
+/*
+Helper functions
+*/
+
+        static string? GetCmdOption(string[] args, string option)
```
trrwilson, line 85: Can/should we use something like `System.CommandLine` for the parsing? It's an extra dependency and not necessarily much less code, but it might be less distracting and more robust code.

trrwilson, line 85: If we do keep the DIY code, I'd suggest putting it in its own small helper file to keep the sample code as focused as possible

trrwilson, line 85: Instance feedback: it's good practice to apply an access modifier (public/private) all the time
Broader/more general: it may be helpful to compile this with static analysis (default rules) turned on to see what it yields -- it'll include things like this instance feedback along with other things like public/private ordering, property/field/method ordering, etc.

scenarios/csharp/dotnet/captioning/Program.cs
```diff
@@ -0,0 +1,417 @@
...
+            return args.Contains (option);
+        }
+
+        static string V2EndpointFromRegion(string region)
+        {
+// Note: Continuous language identification is supported only with v2 endpoints.
+            return "wss://" + region + ".stt.speech.microsoft.com/speech/universal/v2";
+        }
+
+        static string TimestampFromSpeechRecognitionResult(SpeechRecognitionResult result, UserConfig userConfig)
```
trrwilson, line 109: Given the ubiquity and lack of "varying," this likely shouldn't be a variable and should instead be a member or even shared singleton

scenarios/csharp/dotnet/captioning/Program.cs
```diff
@@ -0,0 +1,417 @@
...
+        {
+            DateTime startTime = new DateTime(result.OffsetInTicks);
+            DateTime endTime = startTime.Add(result.Duration);
+            string format;
+
+            if(userConfig.srtEnabled)
```
trrwilson, line 115: Style minutia: space between `if` and opening `(`

scenarios/csharp/dotnet/captioning/Program.cs
```diff
@@ -0,0 +1,417 @@
...
+using Microsoft.CognitiveServices.Speech;
+using Microsoft.CognitiveServices.Speech.Audio;
+
+namespace Caption
+{
+/*
+Types
+*/
+
+    public struct UserConfig
```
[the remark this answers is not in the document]
trrwilson: And this is probably big enough to be its own small helper file with `///` docs for each property

scenarios/csharp/dotnet/captioning/Program.cs
```diff
@@ -0,0 +1,417 @@
...
+            bool profanityFilterRemoveEnabled,
+            bool profanityFilterMaskEnabled,
+            string[]? languageIDLanguages,
+            string? inputFile,
+            string? outputFile,
+            string? phraseList,
+            bool suppressOutputEnabled,
+            bool partialResultsEnabled,
+            string? stablePartialResultThreshold,
+            bool srtEnabled,
```
trrwilson, line 51: A few of these, but especially this one, are not obvious from the name and for sure bear either a longer name (not sure "subripEnabled" is much better, but maybe "captionFileFormatEnabled" or something in that vein) and/or a comment

scenarios/csharp/dotnet/captioning/Program.cs
```diff
@@ -0,0 +1,417 @@
...
+            string format;
+
+            if(userConfig.srtEnabled)
+            {
+// SRT format requires ',' as decimal separator rather than '.'.
+                format = "HH:mm:ss,fff";
+            }
+            else
+            {
+                format = "HH:mm:ss.fff";
```
trrwilson, line 122: Maybe another holy war, but IMO a conditional this simple can/should go into a ternary, e.g.
`format = userConfig.srtEnabled ? "HH:mm:ss,fff" : "HH:mm:ss.fff";`

scenarios/csharp/dotnet/captioning/Program.cs
```diff
@@ -0,0 +1,417 @@
...
+            }
+
+            return String.Format("{0} --> {1}", startTime.ToString(format), endTime.ToString(format));
```
trrwilson, line 125: It's a tradeoff since you can't both assign the format as a variable and use it inside of interpolated strings, but it might be worth using string interpolation for the cleanliness, e.g. remove the above and instead:
```
// SRT caption format requires ',' as the decimal separator instead of '.'
return userConfig.srtEnabled
    ? $"{startTime:HH:mm:ss,fff} --> {endTime:HH:mm:ss,fff}"
    : $"{startTime:HH:mm:ss.fff} --> {endTime:HH:mm:ss.fff}";
```
(My feelings mixed and not as strong on this one since it's not perfect either way)

scenarios/cpp/windows/captioning/caption.cpp
```diff
@@ -0,0 +1,527 @@
...
+/* Simple command line argument parsing functions. See
+https://stackoverflow.com/a/868894
+The value parameter to std::find must be std::string. Otherwise std::find will compare the pointer values rather than the string contents.
+*/
+std::optional<std::string> GetCmdOption(char** begin, char** end, const std::string& option)
+{
+    char** result = std::find(begin, end, option);
+    if (result != end)
+    {
+// We found the option (for example, "-o"), so advance from that to the value (for example, "filename").
```
jhakulin, line 127: In our samples comments are aligned with the indentation. In Caption samples all comments starts from the beginning of line

jhakulin:
Are you going to add Readme.md for each sample similarly we have e.g. in our quickstarts?

scenarios/java/openjdk/captioning/Caption.java
```diff
@@ -0,0 +1,423 @@
...
+    }
+}
+
+public class Caption
+{
+/*
+Helper functions
+*/
+
+    static Optional<String> GetCmdOption(List<String> args, String option)
```
jhakulin, line 119: >GetCmdOption [](http://example.com/codeflow?start=28&length=12)

In Java method names start with lowercase, that has been practice in other Java samples

scenarios/java/openjdk/captioning/Caption.java
```diff
@@ -0,0 +1,423 @@
...
+        this.subscriptionKey = subscriptionKey;
+        this.region = region;
+    }
+}
+
+public class Caption
+{
+/*
+Helper functions
+*/
```
jhakulin, line 117: We do not have similar template in our samples containing following:

/*
Types
*/

/*
Helper functions
*/

/*
Main functions
*/

I would prefer adding high level descriptions for the main methods in the sample and remove these from template.

scenarios/python/console/captioning/caption.py
```diff
@@ -0,0 +1,275 @@
...
+            return None
+    else :
+        return None
+
+def cmd_option_exists(option : str) -> bool :
+    return option in argv
+
+def user_config_from_args() -> Read_Only_Dict :
+    argc = len(argv)
+    if argc < 3 :
```
jhakulin, line 56: Our samples try to follow this Python pep8 convention for blank lines

[https://stackoverflow.com/questions/2953250/python-pep8-blank-lines-convention](https://stackoverflow.com/questions/2953250/python-pep8-blank-lines-convention)

v-jaswel:
> Are you going to add Readme.md for each sample similarly we have e.g. in our quickstarts?

Yes.

scenarios/csharp/dotnet/captioning/Program.cs
```diff
@@ -0,0 +1,417 @@
...
+            }
+
+            if (userConfig.profanityFilterRemoveEnabled)
+            {
+                speechConfig.SetProfanity(ProfanityOption.Removed);
+            }
+            else if (userConfig.profanityFilterMaskEnabled)
+            {
+                speechConfig.SetProfanity(ProfanityOption.Masked);
+            }
```
eric-urban, line 249: @v-jaswel - I've found that the default behavior / value is `Masked`, so we will need to change something here. One thing you could do is let them set the exact string following the flag. For example `-f "Raw"`,  `-f "Removed"`, or  `-f "Masked"`.

scenarios/csharp/dotnet/captioning/Program.cs
```diff
@@ -0,0 +1,417 @@
...
+using System; // File
+using System.Linq; // Enumerable.Contains
+using System.Text; // StringBuilder
+using System.Threading.Tasks; // Task
+
+/* Notes:
+- Install the Speech NuGet package:
+dotnet add package Microsoft.CognitiveServices.Speech
+- The Speech SDK on Windows requires the Microsoft Visual C++ Redistributable for Visual Studio 2019 on the system. See:
+https://docs.microsoft.com/azure/cognitive-services/speech-service/speech-sdk?tabs=windows%2Cubuntu%2Cios-xcode%2Cmac-xcode%2Candroid-studio
```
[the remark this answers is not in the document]
eric-urban: @v-jaswel @trrwilson - Any prerequisites and instructions should probably be moved to the readme file. In the public Carbon repo I see several hundred docs.microsoft.com links, but yes I agree they should be used sparingly and prefer aka.ms.

scenarios/typescript/node/captioning/caption.ts
```diff
@@ -0,0 +1,313 @@
...
+Dependencies
+*/
+
+/* Notes:
+- To install dependencies for this sample:
+npm install microsoft-cognitiveservices-speech-sdk
+- The Speech SDK on Windows requires the Microsoft Visual C++ Redistributable for Visual Studio 2019 on the system. See:
+https://docs.microsoft.com/azure/cognitive-services/speech-service/speech-sdk?tabs=windows%2Cubuntu%2Cios-xcode%2Cmac-xcode%2Candroid-studio
+*/
+import * as sdk from "microsoft-cognitiveservices-speech-sdk";
```
eric-urban, line 13: @v-jaswel - Just a general comment on the JavaScript sample I don't see any dot TS examples in carbon. All of our samples that I found are dot JS.

scenarios/typescript/node/captioning/caption.ts
```diff
@@ -0,0 +1,313 @@
...
+        speechConfig.setProfanity(sdk.ProfanityOption.Removed);
+    }
+    else if (userConfig.profanityFilterMaskEnabled)
+    {
+        speechConfig.setProfanity(sdk.ProfanityOption.Masked);
+    }
+    
+    if (null !== userConfig.stablePartialResultThreshold)
+    {
+        speechConfig.setProperty("SpeechServiceResponse_StablePartialResultThreshold", userConfig.stablePartialResultThreshold);
```
eric-urban, line 188: Use [PropertyId](https://github.com/microsoft/cognitive-services-speech-sdk-js/blob/master/src/sdk/PropertyId.ts#L307) instead of string.
```suggestion
        speechConfig.setProperty(sdk.PropertyId.SpeechServiceResponse_StablePartialResultThreshold, userConfig.stablePartialResultThreshold);
```

scenarios/typescript/node/captioning/caption.ts
```diff
@@ -0,0 +1,313 @@
...
+    }
+    
+    if (userConfig.trueTextEnabled)
+    {
+        speechConfig.setProperty("SpeechServiceResponse_PostProcessingOption", "TrueText");
```
eric-urban, line 193: Use [PropertyId](https://github.com/microsoft/cognitive-services-speech-sdk-js/blob/master/src/sdk/PropertyId.ts#L295) instead of string.

scenarios/csharp/dotnet/captioning/Program.cs
```diff
@@ -0,0 +1,353 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Captioning
+{
+    class Program
+    {
+        private UserConfig userConfig;
+        private const string usage = @"Usage: caption.exe [-f] [-h] [-i path] [-l languages] [-m] [-o path] [-p phrases] [-q] [-r number] [-s] [-t] [-u] <subscriptionKey> <region>
```
eric-urban, line 23: Should it be "captioning.exe"?

scenarios/csharp/dotnet/captioning/UserConfig.cs
```diff
@@ -0,0 +1,75 @@
+//
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE.md file in the project root for full license information.
+//
+
+namespace Captioning
+{
+    public class UserConfig
```
eric-urban, line 8: What are your thoughts about moving this to Program.CS?

scenarios/csharp/dotnet/captioning/Program.cs
```diff
@@ -0,0 +1,353 @@
...
+// dotnet add package Microsoft.CognitiveServices.Speech
+// - The Speech SDK on Windows requires the Microsoft Visual C++ Redistributable for Visual Studio 2019 on the system. See:
+// https://docs.microsoft.com/azure/cognitive-services/speech-service/speech-sdk
+using Microsoft.CognitiveServices.Speech;
+using Microsoft.CognitiveServices.Speech.Audio;
+using System;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
```
eric-urban, line 17: @v-jaswel - The build fails for me without this using statement
// File
using System.IO;

v-jaswel: Thank you, added. Will resolve this when I push.

scenarios/python/console/captioning/captioning.py
```diff
@@ -0,0 +1,248 @@
...
+  LANGUAGE
+    --languages LANG1,LANG2       Enable language identification for specified languages.
+                                  Example: en-US,ja-JP
+
+  INPUT
+    --input FILE                  Input audio from file (default input is the microphone.)
+    --url URL                     Input audio from URL (default input is the microphone.)
+    --format FORMAT               Use compressed audio format.
+                                  Valid only with --file or --url.
+                                  If this is not specified, uncompressed format (wav) is assumed.
```
eric-urban, line 217: What does it mean to assume "wav" when the default is "any"?

scenarios/cpp/windows/captioning/captioning.cpp
```diff
@@ -0,0 +1,538 @@
...
+        // Stops recognition.
+        speechRecognizer->StopContinuousRecognitionAsync().get();
+        
+        return result;
+    }
+};
+
+int main(int argc, char* argv[])
+{
+    const std::string usage = "Usage: caption.exe [...]\n\n"
```
eric-urban, line 485:
```suggestion
    const std::string usage = "Usage: captioning.exe [...]\n\n"
```

scenarios/cpp/windows/captioning/captioning.cpp
```diff
@@ -0,0 +1,538 @@
...
+"  CONNECTION\n"
+"    --key KEY                     Your Azure Speech service subscription key.\n"
+"    --region REGION               Your Azure Speech service region.\n"
+"                                  Examples: westus, eastus\n\n"
+"  LANGUAGE\n"
+"    --languages LANG1,LANG2       Enable language identification for specified languages.\n"
```
eric-urban, line 491: Could this go under the recognition heading? Just double checking is it comma separated or semicolon like phrase list?

scenarios/cpp/windows/captioning/captioning.cpp
```diff
@@ -0,0 +1,538 @@
...
+"                                  Example: en-US,ja-JP\n\n"
+"  INPUT\n"
+"    --input FILE                  Input audio from file (default input is the microphone.)\n"
```
eric-urban, line 494: If if this is supposed to be modeled after the CLI then it should be `--file FILE`

scenarios/cpp/windows/captioning/captioning.cpp
```diff
@@ -0,0 +1,351 @@
...
+
+    std::shared_ptr<UserConfig> m_userConfig = NULL;
+    std::shared_ptr<AudioStreamFormat> m_format = NULL;
+    std::shared_ptr<BinaryFileReader> m_callback = NULL;
+    std::shared_ptr<PullAudioInputStream> m_stream = NULL;
+
+    std::string TimestampFromSpeechRecognitionResult(std::shared_ptr<SpeechRecognitionResult> result)
+    {
+        std::ostringstream strTimestamp;
+        auto timestamp = TimestampFromTicks(result->Offset(), result->Offset() + result->Duration());
```
dargilco, line 40: I know folks have different opinions, but I would never use "auto" in any sample code. I actually don't use it in any code I know someone else will need to work on (which is everything we do at work). Using "auto" makes the code unreadable. We want sample code to be readable.

dargilco: Don't remember if I already commented on this...

dargilco (approved):

scenarios/cpp/windows/captioning/caption.cpp
```diff
@@ -0,0 +1,527 @@
...
+
+int main(int argc, char* argv[])
+{
+    const std::string usage = "Usage: caption.exe [-f] [-h] [-i file] [-l languages] [-m] [-o file] [-p phrases] [-q] [-r number] [-s] [-t] [-u] <subscriptionKey> <region>\n"
+    "              -f: Enable profanity filter (remove profanity). Overrides -m.\n"
+    "              -h: Show this help and stop.\n"
+    "              -i: Input audio file *file* (default input is from the microphone.)\n"
+    "    -l languages: Enable language identification for specified *languages*.\n"
+    "                  Example: en-US,ja-JP\n"
+    "              -m: Enable profanity filter (mask profanity). -f overrides this.\n"
```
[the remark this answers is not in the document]
v-jaswel: -f and -m are now replaced by --profanity [raw|mask|remove] (default mask)

scenarios/cpp/windows/captioning/caption.cpp
```diff
@@ -0,0 +1,527 @@
...
+    speechRecognizer->StopContinuousRecognitionAsync().get();
+    
+    return result;
+}
+
+int main(int argc, char* argv[])
+{
+    const std::string usage = "Usage: caption.exe [-f] [-h] [-i file] [-l languages] [-m] [-o file] [-p phrases] [-q] [-r number] [-s] [-t] [-u] <subscriptionKey> <region>\n"
+    "              -f: Enable profanity filter (remove profanity). Overrides -m.\n"
+    "              -h: Show this help and stop.\n"
```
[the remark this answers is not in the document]
v-jaswel: Good point, will do this in new usage string.

scenarios/cpp/windows/captioning/caption.cpp
```diff
@@ -0,0 +1,527 @@
...
+    "              -i: Input audio file *file* (default input is from the microphone.)\n"
+    "    -l languages: Enable language identification for specified *languages*.\n"
+    "                  Example: en-US,ja-JP\n"
+    "              -m: Enable profanity filter (mask profanity). -f overrides this.\n"
+    "         -o file: Output to *file*.\n"
```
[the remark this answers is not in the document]
v-jaswel: Will do.

scenarios/cpp/windows/captioning/caption.cpp
```diff
@@ -0,0 +1,527 @@
...
+    "      -p phrases: Add specified *phrases*.\n"
+    "                  Example: Constoso;Jessie;Rehaan\n"
+    "              -q: Suppress console output (except errors).\n"
+    "       -r number: Set stable partial result threshold to *number*.\n"
+    "                  Example: 3\n"
+    "              -s: Output captions in SRT format (default is WebVTT format.)\n"
+    "              -t: Enable TrueText.\n"
```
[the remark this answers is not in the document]
v-jaswel: TrueText is now on all the time (i.e. no longer a command line flag).

scenarios/cpp/windows/captioning/captioning.cpp
v-jaswel: Travis suggested using "var" when the type can be inferred from the right hand side (he did mention it might be a holy war : )
That said, in this case, the type can't be inferred, so I will replace "auto" here and look for other cases like this.

scenarios/cpp/windows/captioning/captioning.cpp
v-jaswel: True but we also have possible output to a file. I could use --file for input and --output for output but that feels asymmetric.
I would use --output file FILE but the command line arg parsing in the sample only supports one param to each flag. --output_file FILE seems like a pain for the user to type.

scenarios/cpp/windows/captioning/captioning.cpp
v-jaswel: Good point. I thought the service expected a comma delimited list but no, it wants an array, so we can use any delimiter. I'll use ; instead to match SPX.

scenarios/python/console/captioning/captioning.py
v-jaswel: Yeah that is confusing. If you don't use --format at all, it assumes .wav. If you do use --format but don't provide a value, it assumes "any". I'll try to clarify.

dargilco (dismissed):

dargilco (approved):

dargilco (approved):

PR #1462, opened 2022-04-11
latest C# quickstart for Docs

eric-urban:
## Purpose
Request to add the latest C# STT quickstart. Then we can import to Docs via the <stt-quickstart> tag.

## Pull Request Type
Sample is created via dotnet CLI without a VS solution.

```
[ ] Bugfix
[ ] Feature
[x ] Documentation content changes
[ ] Other... Please describe:
```


eric-urban:
@dargilco @trrwilson please check

quickstart/csharp/dotnetcore/from-microphone/README.md
```diff
@@ -2,72 +2,5 @@
 
 This sample demonstrates how to recognize speech with C# under .NET Core 3.1 (Windows, macOS, or Linux) using the Speech SDK.
 
-> **Note:**
-> We currently only support [specific Linux distributions and target architectures](https://docs.microsoft.com/azure/cognitive-services/speech-service/speech-sdk?tabs=linux).
+For instructions on how to create and run this console application using the Speech SDK, see the [Speech to text quickstart](https://docs.microsoft.com/en-us/azure/cognitive-services/speech-service/get-started-speech-to-text?pivots=programming-language-csharp) documentation.
```
dargilco, line 5: Remove "en-us" from the link

quickstart/csharp/dotnetcore/from-microphone/helloworld.sln
```diff
@@ -1,25 +0,0 @@
-
```
dargilco, line 1: Please don't delete the .sln file. Carbon automaticly builds the samples using the VS solution file in our CI/CD pipeline. And I think we want users to have VS solutions for all Windows samples, even if you can build on the command line. So we need to find a way to document both options (VS build or command line build). So maybe leave the VS and updat the README.md file to mention both options.

quickstart/csharp/dotnetcore/from-microphone/helloworld/Program.cs
```diff
@@ -3,64 +3,51 @@
 // Licensed under the MIT license. See LICENSE.md file in the project root for full license information.
 //
 
-// <code>
+// <stt-quickstart>
```
dargilco, line 6: I was thinking maybe all our tags should start with a common prefix such as "code-embedded-in-microsoft-docs-..." or something of the sort, so both we (Carbon devs) and external developers know what these are for. Maybe discussed in the next meeting (but I'm out on Monday). Don't let that block this check in, just something to discuss. I think that would be helpful.

quickstart/csharp/dotnetcore/from-microphone/helloworld/Program.cs
```diff
@@ -3,64 +3,51 @@
...
+    static string YourSubscriptionKey = "YourSubscriptionKey";
+    static string YourServiceRegion = "YourServiceRegion";
+
+    async static Task Main(string[] args)
     {
-        public static async Task RecognizeSpeechAsync()
-        {
-            // Creates an instance of a speech config with specified subscription key and service region.
-            // Replace with your own subscription key // and service region (e.g., "westus").
-            var config = SpeechConfig.FromSubscription("YourSubscriptionKey", "YourServiceRegion");
```
dargilco, line 19: Please bring back the comments. We did not agree on deleting comments from GitHub quickstarts. If/when we have two versions of quickstarts, one solely dedicated to oneline-docs, we can remove comments from that version. But as long as we only have one version, we need comments in it.

dargilco (changes requested):

PR #1502, opened 2022-05-20
[Ingestion Client] Switch to Text Analytics async API, use Text Analytics SDK

HenryvanderVegte:
## Purpose
Switches from old text analytics requests (custom HTTP results, immediate results) to the async text analytics API (submit job, wait for completion, get results) through the Text Analytics SDK

## Pull Request Type

```
[ ] Bugfix
[x] Feature
[ ] Other... Please describe:
```


PanosPeriorellis (approved):
The thing i am not sure about is whether TA would choke with large number of requests. Potentially we can generate a lot of traffic for them.

Also we cannot push all utts from a call in a single request. this needs to be managed. they have a document size that can help us determine the number of utts they accept per request. did not see that in the code.

PanosPeriorellis (approved):

PR #1520, opened 2022-06-07
[IngestionClient] Catch SocketTimeoutException + Fix Text Analytics when using either PiiRedaction or SentimentAnalysis

HenryvanderVegte:
## Purpose
- Catch SocketTimeoutException in BatchClient
- Fix bug in Text Analytics where Pii/Sentiment did not get used if only one of the two options was picked

## Pull Request Type

```
[x] Bugfix
[ ] Feature
[ ] Other... Please describe:
```


samples/ingestion/ingestion-client/Connector/BatchClient.cs
```diff
@@ -118,6 +119,13 @@ private static async Task<HttpResponseMessage> SendHttpRequestMessage(HttpReques
 
                 return responseMessage;
             }
+            catch (HttpRequestException requestException)
+                when (requestException.InnerException != null &&
+                requestException.InnerException is SocketException socketException &&
+                socketException.SocketErrorCode == SocketError.TimedOut)
```
shahidkhuram, line 125: Ideally retry on any socker error, right? You can't really call it a TimeoutException anymore but I do think we should handle this broader category of failures and attempt retry

HenryvanderVegte: I don't think we should retry in all other cases as most of them are not transient, so no need to retry there. 
I think I'd prefer to only check for the timeout for now, and add other cases if needed later.

PanosPeriorellis (dismissed):

PanosPeriorellis (dismissed):

PR #1574, opened 2022-07-14
Add call center C# scenario

v-jaswel:
## Purpose
Add call center C# scenario

## Pull Request Type
```
[ ] Bugfix
[ ] Feature
[x] Other... Please describe: Add call center C# scenario
```


scenarios/csharp/dotnetcore/call-center/call-center/Program.cs
```diff
@@ -0,0 +1,408 @@
...
+            {
+                Console.WriteLine($"Waiting {waitSeconds} seconds for conversation analysis to complete.");
+                Thread.Sleep(waitSeconds * 1000);
+                done = await GetConversationAnalysisStatus(conversationAnalysisUrl);
+            }
+            Console.WriteLine();
+            return;
+        }
+
+        private async Task<((string, string)[], (string, string)[][])> GetConversationAnalysis(string conversationAnalysisUrl)
```
trrwilson, line 268: >((string, string)[], (string, string)[][]) [](http://example.com/codeflow?start=27&length=42)

This is a sophisticated enough type that it merits strong representation -- the summary and analysis concepts could each have a class. "A tuple of arrays of tuples of two strings and a two-dimensional matrix of tuples of two strings" just doesn't flow off the tongue :D

scenarios/csharp/dotnetcore/call-center/call-center/Program.cs
```diff
@@ -0,0 +1,408 @@
...
+            try
+            {
+                if (args.Contains("--help"))
+                {
+                    Console.WriteLine(usage);
+                }
+                else
+                {
+                    Program program = new Program(args, usage);
+                    program.Run().Wait();
```
trrwilson, line 399: >Wait [](http://example.com/codeflow?start=34&length=4)

I believe C# 8.0 (default for .NET Core 3.1) should allow you to make `Main` an `async Task Main` instead of `void Main` -- and use `await` directly. 

That might also make it convenient to just treat `Run` as a static method (to avoid mixing/matching static and instance concepts; an instance "Program" constructor is atypical).

scenarios/csharp/dotnetcore/call-center/call-center/RestHelper.cs
```diff
@@ -0,0 +1,87 @@
...
+using System.Net;
+using System.Net.Http;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace CallCenter
+{
+    public static class RestHelper
+    {
+        public static async Task<(HttpResponseMessage, string)> SendGet(string uri, string key, HttpStatusCode[] expectedStatusCodes)
```
trrwilson, line 17: >HttpResponseMessage [](http://example.com/codeflow?start=34&length=19)

C# 7+ should allow you to "name" Tuple constituents, e.g. `Task<(HttpResponseMessage ResponseMessage, string OtherThing)>` -- which would then let you consume `.ResponseMessage` instead of `.Item1` and the like. Super helpful with non-obvious tuples, as is here--I'm actually still not sure what `response_2` (and thus `.Item2`) actually is!

scenarios/csharp/dotnetcore/call-center/call-center/Program.cs
v-jaswel: Thank you for your comments! I created a new type (ConversationAnalysisResult) for this.

scenarios/csharp/dotnetcore/call-center/call-center/Program.cs
v-jaswel: Thank you, I converted Main to public async static Task and got rid of the Program constructor. However, because I have an instance of UserConfig as a member of the Program class, I get the warning:

C:\Users\v-jasonwells\Documents\backup\code\call_center_20220419\cs\call-center\Program.cs(31,28): warning CS8618: Non-nullable field 'userConfig' must contain a non-null value when exiting constructor. Consider declaring the field as nullable. [C:\Users\v-jasonwells\Documents\backup\code\call_center_20220419\cs\call-center\call-center.csproj]

That happens even though I initialize it while new-ing Program:
```
new Program { userConfig = new UserConfig(args, usage) }
```

I'd rather not make userConfig nullable because then I have to null-check it every time I use it, unless that's the standard practice.

scenarios/csharp/dotnetcore/call-center/call-center/RestHelper.cs
v-jaswel: Thanks, that feels more like F# records which is nice. I did this and got rid of all the .Item1/.Item2s.

dargilco (approved):

dargilco (approved):

PR #1577, opened 2022-07-18
Integration of Conversations PII API with ingestion client

bidisha-c:
## Purpose

* Integrating the conversational PII preview API with ingestion client to provide clients opportunity to integrate with the feature.

## Pull Request Type

```
[ ] Bugfix
[x] Feature
[ ] Other... Please describe:
```


samples/ingestion/ingestion-client/FetchTranscription/Language/AnalyzeConversationsProvider.cs
```diff
@@ -0,0 +1,217 @@
...
+            {
+                DisplayName = "IngestionClient",
+                AnalysisInput = new AnalysisInput(new[]
+                {
+                    new Conversation
+                    {
+                        Id = Guid.NewGuid().ToString(),
+                        Language = Locale,
+                        Modality = Modality.transcript,
+                        ConversationItems = speechTranscript.RecognizedPhrases.Where(r => !string.IsNullOrEmpty(r.NBest.First().Lexical))
```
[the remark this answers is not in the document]
bidisha-c: Yes I plan to add the chunking as part of the next iteration of the PR. we are also working in parallel to support longer documents in TA

samples/ingestion/ingestion-client/Connector/Serializable/Conversations/RedactedTranscriptContent.cs
```diff
@@ -0,0 +1,63 @@
...
+        {
+            get;
+            set;
+        }
+
+        /// <summary>
+        /// Gets  Redacted audio timing information for speech input.
+        /// </summary>
+        [JsonProperty(PropertyName = "redactedAudioTimings")]
+        public IList<AudioTiming> RedactedAudioTimings
```
shahidkhuram, line 57: In the current output Json format, we are losing a lot of detail/fidelity from the language API response because we are just redacting the text directly. Instead of patching the speech response I think we should add the language service response in parallel to the JSON so we don't have loss of information (e.g. customer needed original as well as redacted text, or redaction timings, pii categories etc). 

IMO, we should stick with the documented API responses from the respective services and ADD any helper fields to make consumption easier if/when needed. If the language service responses have some conflicting/confusing names as compared to the speech response then we can look at transforms but let's start with the existing API responses.

Bidisha, it would be great if you can propose an update to the ingestion client output JSON format that will retain all details and works for PII, summary and any other insights we want to add in the future. Happy to help brainstorm as needed.

bidisha-c: One option is adding the response directly to the SpeechTranscript object. The output will look something like this. @shahidkhuram @HenryvanderVegte feel free to add your comments here:

````
{
  "source": "https://<sourceurl>",
  "timeStamp": "2022-07-21T19:12:04Z",
  "durationInTicks": 116400000,
  "duration": "PT11.64S",
  "combinedRecognizedPhrases": [
    {
      "channel": 0,
      "lexical": "my name is jane doe",
      "itn": "my name is jane doe",
      "maskedITN": "",
      "display": "My name is jane doe."
    }
  ],
  "recognizedPhrases": [
    {
      "recognitionStatus": "Success",
      "channel": 0,
      "speaker": 0,
      "offset": "PT0.13S",
      "duration": "PT10.77S",
      "offsetInTicks": 1300000,
      "durationInTicks": 107700000,
      "nBest": [
        {
          "confidence": 0.8295387,
          "lexical": "my name is jane doe",
          "itn": "my name is jane doe",
          "maskedITN": "",
          "display": "My name is jane doe.",
          "sentiment": {
            "negative": 0,
            "neutral": 0,
            "positive": 0
          },
          "words": []
        }
      ]
    }
  ],
  "conversationAnalyticsResults": {
    "PiiResults": {
      "conversations": [
        {
          "id": "0",
          "conversationItems": [
            {
              "id": "0",
              "redactedContent": {
                "itn": "my name is ********",
                "lexical": "my name is ********",
                "text": "My name is ********."
              },
              "entities": [
                {
                  "text": "jane doe",
                  "category": "Name",
                  "offset": 11,
                  "length": 8,
                  "confidenceScore": 0.81
                }
              ]
            }
          ],
          "warnings": []
        }
      ],
      "errors": []
    },
    "summarizationResults": {
      "conversations": [
        {
          "id": "0",
          "summaries": [],
          "warning": []
        }
      ],
      "errors": []
    }
  }
}

File: samples/ingestion/ingestion-client/FetchTranscription/Language/AnalyzeConversationsProvider.cs (hunk 1 above)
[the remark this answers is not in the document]
Comment (v-jaswel) [reply samples/ingestion/ingestion-client/FetchTranscription/Language/AnalyzeConversationsProvider.cs:67]: I don't know if you'll run into this issue or not. To pass the automated checking when submitting a sample to this repo, I had to have my .csproj target .NET Core 3.1 instead of .NET 6. As a result, the Enumerable.Chunk function wasn't available to me. However, you can get the source for Chunk here:
https://github.com/dotnet/runtime/blob/main/src/libraries/System.Linq/src/System/Linq/Chunk.cs

I had to make a couple of tweaks to get it to work with C# 8, so my revised version is here:
https://github.com/Azure-Samples/cognitive-services-speech-sdk/blob/v-jaswel-patch-3/scenarios/csharp/dotnetcore/call_center/call_center/Chunk.cs

Let me know if I can answer any questions.

File: samples/ingestion/ingestion-client/Connector/Serializable/Conversations/RedactedTranscriptContent.cs (hunk 2 above)
Comment (HenryvanderVegte) [reply samples/ingestion/ingestion-client/Connector/Serializable/Conversations/RedactedTranscriptContent.cs:57]: I thought removing the Pii completely from the transcript is intended? 
Please check with Panos as he got more customer insights regarding this.

Comment (HenryvanderVegte) [reply samples/ingestion/ingestion-client/Connector/Serializable/Conversations/RedactedTranscriptContent.cs:57]: what's the id here? 
does it correspond with the order of items in the recognizedPhrases?

File: samples/ingestion/ingestion-client/FetchTranscription/FetchTranscriptionEnvironmentVariables.cs [hunk 3]
```diff
@@ -59,5 +66,26 @@ public static class FetchTranscriptionEnvironmentVariables
...
+                {
+                    ConversationPiiCategory.All
+                };
+            }
+
+            var piiCategoriesList = piiCategories.Split(',').ToList();
+            var categories = new List<ConversationPiiCategory>();
+            piiCategoriesList.ForEach(category =>
+            {
+                Enum.TryParse(category, ignoreCase: true, out ConversationPiiCategory catgoryEnum);
```
HenryvanderVegte, line 84: if TryParse would return false I believe categoryEnum would be null, so we shouldn't add it

HenryvanderVegte, line 84: typo "catgoryEnum"

samples/ingestion/ingestion-client/FetchTranscription/FetchTranscription.csproj
```diff
@@ -16,6 +16,7 @@
     <OutputPath>bin\Release\</OutputPath>
   </PropertyGroup>  
   <ItemGroup>
+    <PackageReference Include="Azure.AI.Language.Conversations" Version="1.1.0-beta.1" />
```
HenryvanderVegte, line 19: can't we get all of the classes in Connector/Serializable/Conversations also from this package here as well?

samples/ingestion/ingestion-client/FetchTranscription/Language/AnalyzeConversationsProvider.cs
```diff
@@ -0,0 +1,255 @@
...
+        {
+            ConversationAnalysisClient = new ConversationAnalysisClient(new Uri($"https://{region}.api.cognitive.microsoft.com"), new AzureKeyCredential(subscriptionKey));
+
+            Locale = locale;
+            Log = log;
+        }
+
+        public bool IsConversationalPiiEnabled()
+        {
+            return Locale.Contains("en", StringComparison.OrdinalIgnoreCase) && FetchTranscriptionEnvironmentVariables.ConversationPiiSetting != Connector.Enums.ConversationPiiSetting.None;
```
HenryvanderVegte, line 45: Keep in mind that you can't update the ingestion client deployments, so customers would need to re-deploy their solution if you're planning to add new locales.
Maybe it'd be better to just let the TA request fail?

samples/ingestion/ingestion-client/FetchTranscription/Language/AnalyzeConversationsProvider.cs
```diff
@@ -0,0 +1,255 @@
...
+                {
+                    new Conversation
+                    {
+                        Id = Guid.NewGuid().ToString(),
+                        Language = Locale,
+                        Modality = Modality.transcript,
+                        ConversationItems = speechTranscript.RecognizedPhrases.Where(r => !string.IsNullOrEmpty(r.NBest.First().Lexical))
+                            .Select(item => new ConversationItem()
+                            {
+                                Text = item.NBest.First().Display,
```
HenryvanderVegte, line 70: nit: you could get item.NBest.First() once in the select, then you don't need to call First() so many times

samples/ingestion/ingestion-client/Connector/Serializable/Conversations/RedactedTranscriptContent.cs
bidisha-c: for conversation item id: we are using the $"{item.Channel}_{item.Offset}
for conversation id we are using a GUID as it just the identifier for the whole conversation and do not want to tie it to any channel.

samples/ingestion/ingestion-client/FetchTranscription/Language/AnalyzeConversationsProvider.cs
```diff
@@ -0,0 +1,255 @@
...
+                var jobResults = result.RootElement;
+                var jobId = jobResults.GetProperty("jobId");
+
+                return (new List<string> { jobId.GetString() }, errors);
+            }
+            catch (OperationCanceledException)
+            {
+                throw new TimeoutException($"The operation has timed out after {RequestTimeout.TotalSeconds} seconds.");
+            }
+
```
HenryvanderVegte, line 206: remove new line here

samples/ingestion/ingestion-client/FetchTranscription/FetchTranscriptionEnvironmentVariables.cs
bidisha-c: fixed :-)

samples/ingestion/ingestion-client/FetchTranscription/FetchTranscription.csproj
HenryvanderVegte: also, I don't think we should have a "-beta.1" version in here

samples/ingestion/ingestion-client/Setup/ArmTemplateBatch.json
```diff
@@ -947,7 +960,8 @@
                 "ConsolidatedFilesOutputContainer": "[variables('ConsolidatedFilesOutputContainer')]",
                 "CreateAudioProcessedContainer": "[variables('CreateAudioProcessedContainer')]",
                 "AudioProcessedContainer": "[variables('AudioProcessedContainer')]",
-                "PiiCategories": "[variables('PiiCategories')]"
+                "PiiCategories": "[variables('PiiCategories')]",
```
HenryvanderVegte, line 963: are all categories completely identical for both TA endpoints?

samples/ingestion/ingestion-client/Setup/ArmTemplateBatch.json
```diff
@@ -204,6 +204,18 @@
                 "description": "A value indicating whether personally identifiable information (PII) redaction is requested. Will only be performed if a Text Analytics Key and Region is provided."
             }
         },
+        "ConversationPiiRedaction": {
+            "defaultValue": "None",
+            "type": "String",
+            "allowedValues": [
+                "None",
+                "DoNotIncludeAudioRedaction",
+                "IncludeAudioRedaction"
```
HenryvanderVegte, line 213: IncludeAudioRedaction does not redact audio yet, or?
I'd hide it to avoid confusion

samples/ingestion/ingestion-client/FetchTranscription/FetchTranscription.csproj
bidisha-c: The Azure.AI.Language.Conversations SDK is auto-generated using the cadl and it does not return the classes.

samples/ingestion/ingestion-client/FetchTranscription/Language/AnalyzeConversationsProvider.cs
bidisha-c: thats fair.

samples/ingestion/ingestion-client/Connector/Enums/ConversationPiiCategory.cs
```diff
@@ -0,0 +1,18 @@
+// <copyright file="ConversationPiiCategory.cs" company="Microsoft Corporation">
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE.md file in the project root for full license information.
+// </copyright>
+
+namespace Connector.Enums
```
HenryvanderVegte, line 6: imo it should be up to TA to reject requests if there's an unexpected category. 
If you add new entities, you won't be able to pick them up without redeploying the ingestion client

samples/ingestion/ingestion-client/Setup/ArmTemplateBatch.json
bidisha-c: right

samples/ingestion/ingestion-client/FetchTranscription/Language/AnalyzeConversationsProvider.cs
```diff
@@ -0,0 +1,255 @@
+﻿// <copyright file="AnalyzeConversationsProvider.cs" company="Microsoft Corporation">
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE.md file in the project root for full license information.
+// </copyright>
+
+namespace FetchTranscription.Language
```
HenryvanderVegte, line 6: not sure if the distinction between Language and TextAnalytics makes sense. 
I think the logic in here is very similar to the one in TextAnalyticsProvider, is it possible to merge them?

samples/ingestion/ingestion-client/Connector/Enums/ConversationPiiCategory.cs
bidisha-c: I issue I have is we cannot surface these errors when the ARM template is deploying. Is there a way to run that as validations when we are deploying?

samples/ingestion/ingestion-client/FetchTranscription/Language/AnalyzeConversationsProvider.cs
bidisha-c: I actually had the ConversationAnalysisProvider part of the provider class but moved it out for a couple of reasons:
1. They are separate APIs
2. TA is specifically for text/document capabilities and the new conversations suites are for supporting conversations. The distinction is important as they support different input modalities. In the future I think for speech specifically, all capabilities will need to migrate to the conversation capabilities rather than the TA capabilities as and when we launch more features. 

That said, if it still makes more sense to have them together for a better customer experience, I am happy to merge them.

samples/ingestion/ingestion-client/FetchTranscription/Language/AnalyzeConversationsProvider.cs
bidisha-c: fixed

samples/ingestion/ingestion-client/FetchTranscription/Language/AnalyzeConversationsProvider.cs
bidisha-c: there is comment below so need the new line here :-)

samples/ingestion/ingestion-client/Setup/ArmTemplateBatch.json
bidisha-c: no. But do we want to have a separate variable for pii and conversation pii for customers?

samples/ingestion/ingestion-client/Setup/ArmTemplateBatch.json
bidisha-c: For now just having an enable only. If we expose the TA outputs directly, we can enable it by default.

samples/ingestion/ingestion-client/Connector/Enums/ConversationPiiCategory.cs
HenryvanderVegte: I think I searched for something like this a while ago and couldn't find anything :(

samples/ingestion/ingestion-client/Setup/ArmTemplateBatch.json
HenryvanderVegte: Makes sense to use the variable for both, was just curious :D

samples/ingestion/ingestion-client/Connector/Serializable/Conversations/RedactedTranscriptContent.cs
shahidkhuram: @bidisha-c {item.Channel}_{item.Offset} is a bit confusing. How come channel and offset/id are not unique fields on the response? Alternatively each channel can be a separate list if you want to drop the offset.

@HenryvanderVegte, I believe you said you had a proposal you would share an example for on how to merge this at the segment level? Could you please share? Ideally also share an example on how that would work when language insights apply to multiple segments, or don't have a clear correlation (e.g. summarization)

bidisha-c: As long as Offset is unique, we do not really need the channel here. I can update to have the Offset only.

samples/ingestion/ingestion-client/Setup/ArmTemplateBatch.json
bidisha-c: This is enabled now.

samples/ingestion/ingestion-client/FetchTranscription/FetchTranscriptionEnvironmentVariables.cs
bidisha-c: Removed the enum as discussed in a separate thread, we do not want to declare categories as enum as that will require clients to re-deploy the ARM template when we add more PII types.

samples/ingestion/ingestion-client/Setup/ArmTemplateBatch.json
bidisha-c: Keeping it as a separate variable but feel free to update this.

samples/ingestion/ingestion-client/FetchTranscription/Language/AnalyzeConversationsProvider.cs
bidisha-c: Keeping it separate. Going forward the TA provider will be deprecated in favor of the ConversationAnalysisProvider

samples/ingestion/ingestion-client/Connector/Enums/ConversationPiiCategory.cs
bidisha-c: As discussed removed this enum for now.

samples/ingestion/ingestion-client/FetchTranscription/TextAnalytics/TextAnalyticsProvider.cs
```diff
@@ -251,31 +269,34 @@ public async Task<bool> TextAnalyticsRequestsCompleted(IEnumerable<AudioFileInfo
...
+                    var piiResult = piiResults.Where(document => document.Id.Equals($"{channel}", StringComparison.OrdinalIgnoreCase)).SingleOrDefault();
+                    if (piiResult != null)
+                    {
+                        var redactedText = piiResult.Entities.RedactedText;
 
-                    combinedRecognizedPhrase.Display = redactedText;
-                    combinedRecognizedPhrase.ITN = string.Empty;
-                    combinedRecognizedPhrase.MaskedITN = string.Empty;
-                    combinedRecognizedPhrase.Lexical = string.Empty;
+                        combinedRecognizedPhrase.Display = redactedText;
```
HenryvanderVegte, line 279: are there tabs instead of spaces here?
highlighting in gh looks weird.

HenryvanderVegte:
Please merge with master to have this commit 
https://github.com/Azure-Samples/cognitive-services-speech-sdk/commit/ad52428dd0e5e1fc036c27043911775d0c0e6014
as part of your PR

samples/ingestion/ingestion-client/Connector/Serializable/Conversations/RedactedTranscriptContent.cs
HenryvanderVegte: @bidisha-c can you update the example here to reflect your changes?

I agree with Khuram that {item.Channel}_{item.Offset}  is a bit confusing. We could also just have a channelIndex as part of the json. Even if it is not part of the original conversationAnalyticsResults, it should scale if we just add it everywhere like:

`"conversationAnalyticsResults": {
    "PiiResults": {
      "conversations": [
        {
          "id": "0",
          "conversationItems": [
            {
              "id": "PT0.13S",
              **"channel": "0",**
              "redactedContent": {
                "itn": "my name is ********",
                "lexical": "my name is ********",
                "text": "My name is ********."
              },
              "entities": [
                {
                  "text": "jane doe",
                  "category": "Name",
                  "offset": 11,
                  "length": 8,
                  "confidenceScore": 0.81
                }
              ]
            }
          ],
          "warnings": []
        }
      ],
      "errors": []
    },
    "summarizationResults": {
      "conversations": [
        {
          "id": "0",
          "summaries": [],
          "warning": []
        }
      ],
      "errors": []
    }
  }`

HenryvanderVegte (dismissed):

samples/ingestion/ingestion-client/Connector/Serializable/Conversations/RedactedTranscriptContent.cs
HenryvanderVegte: or if you want to keep the format, please add some documentation in code and in the readme since the id format is not intuitive imo

bidisha-c: ID is the field we use in TA for uniquely identifying conversation turns which is missing speech output. I could not find a unique identiofier for each recognized phrase. The customer can disregard it as it we are augmenting the TA output to have the offset and channel to map to the speech response. If this is still confusing we can remove the ID field itself.

samples/ingestion/ingestion-client/FetchTranscription/Language/AnalyzeConversationsProvider.cs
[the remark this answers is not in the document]
bidisha-c: Currently not planning to handle chunking as TA async Conversation API will start supporting longer conversations.

samples/ingestion/ingestion-client/Connector/Serializable/Conversations/RedactedTranscriptContent.cs
bidisha-c: I have addressed most of the comments as part of this commit. 

1. Support for chunking
2. Support for combinedRedactedContent
3. Replacing the conversation item ID with an index in response to simplify for customer.
Here is a sample output 

```
"conversationAnalyticsResults":` {
    "conversationPiiResults": {
      "combinedRedactedContent": [
        {
          "channel": "1",
          "lexical": "hi how can I help you",
          "itn": "hi how can I help you",
          "display": "Hi how can I help you?"
        },
        {
          "channel": "1",
          "lexical": "my name is ********",
          "itn": "my name is ********",
          "display": "My name is ********."
        }
      ],
      "conversations": [
        {
          "conversationItems": [
            {
              "id": "1",
              "channel": "0",
              "offset": "PT0.15S",
              "redactedContent": {
                "itn": "hi how can I help you",
                "lexical": "hi how can I help you",
                "text": "hi how can I help you?"
              },
              "entities": []
            },
			{
              "id": "2",
              "channel": "1",
              "offset": "PT0.35S",
              "redactedContent": {
                "itn": "my name is ********",
                "lexical": "my name is ********",
                "text": "My name is ********."
              },
              "entities": [
                {
                  "Text": "Jane Doe",
                  "Offset": 11,
                  "Length": 8,
                  "Score": 0.0,
                  "Category": "Name"
                }
              ]
            }
          ],
          "warnings": []
        }
      ]
    }
  }
```

HenryvanderVegte: Looks good! Thanks!

HenryvanderVegte (approved):

PR #1614, opened 2022-08-12
[IngestionClient] Fix StartTranscriptionHelper service bus message completion bug

HenryvanderVegte:
## Purpose
There's was a bug introduced here
https://github.com/Azure-Samples/cognitive-services-speech-sdk/commit/74c104ab5b102c8d41270f1569fe2ea88b5cbf78
that would result in most service bus messages processed by the StartTranscriptionFunction not being completed (as only the nth message of each chunk would be completed).

This PR fixes it by completing all processed messages.

We didn't notice that bug earlier, since we by default have a maxDeliveryCount of 1 on start transcription service bus messages - so the additional messages would get dead lettered. However, if the maxDeliveryCount is being increased, it would lead to the same service bus message being processed twice.

## Pull Request Type

```
[x] Bugfix
[ ] Feature
[ ] Other... Please describe:
```


mahilleb-msft (dismissed):
:shipit:

mahilleb-msft (approved):
:shipit:

PR #1695, opened 2022-10-17
[IngestionClient] Fix http request retry, text analytics NRE and prepare for same service bus message on retry

HenryvanderVegte:
## Purpose
* Fix bug: Http request retry in Polly was trying to re-use the same HttpRequest - needs to be recreated
* Fix bug: We might run into NRE exceptions if we got only AudioLevelRequests OR UtteranceLevelRequests
* Prepare scenario in which the same audio file is processed twice (which might happen if a message lock is lost, for instance if the function is restarted during processing - or if the audio file is deleted and then re-uploaded).

## Pull Request Type

```
[x] Bugfix
[ ] Feature
[ ] Other... Please describe:
```


mahilleb-msft (approved):
:shipit:

PR #1775, opened 2022-12-07
remove .npmrc files that block "npm install" by asking for login

glharper:
## Purpose

* ...

## Pull Request Type

```
[ x] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

### DIFF
--- a/quickstart/javascript/node/conversation-transcription/.npmrc
+++ b/quickstart/javascript/node/conversation-transcription/.npmrc
@@ -1,4 +1,0 @@
-# Auto generated file from Gardener Plugin CentralFeedServiceAdoptionPlugin
-registry=https://pkgs.dev.azure.com/msasg/skyman/_packaging/skyman_PublicPackages/npm/registry/
-
-always-auth=true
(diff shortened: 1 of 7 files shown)


dargilco (approved):

quickstart/javascript/node/conversation-transcription/.npmrc
```diff
@@ -1,4 +0,0 @@
-# Auto generated file from Gardener Plugin CentralFeedServiceAdoptionPlugin
```
BrianMouncer, line 1: These were all added by the SDL security team.
I don't think you can remove these.

Well, technically I think you can reduce them to one .npmrc file in the root of the enlistment, but that won't get rid of the user sign-in prompt when building locally...

Was there some other reason to remove these build config file?

BrianMouncer: aka, if you remove these npmrc files, you will get a CI build failure that the node project is pulling packages from an untrusted package source...

glharper: The samples in the [JS azure SDK](https://github.com/Azure-Samples/azure-sdk-for-js-samples) manages to present their samples without a .npmrc. Is there any way we can get the same latitude?

PR #1829, opened 2023-02-10
Updating sample code to use latest API versions

ut-karsh:
## Purpose
* Only updated C# in previous PR. Updating other programming languages in this PR to Speech REST API version 3.1 and ConversationAnalysis API version 2022-10-01-preview.

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
[ x] Other... Please describe: Updating API versions, no functional changes.
```

### DIFF
--- a/scenarios/python/console/call-center/call_center.py
+++ b/scenarios/python/console/call-center/call_center.py
@@ -25,7 +25,7 @@ SENTIMENT_ANALYSIS_PATH = "/language/:analyze-text";
 SENTIMENT_ANALYSIS_QUERY = "?api-version=2022-05-01";
 CONVERSATION_ANALYSIS_PATH = "/language/analyze-conversations/jobs";
 CONVERSATION_ANALYSIS_QUERY = "?api-version=2022-10-01-preview";
-CONVERSATION_SUMMARY_MODEL_VERSION = latest;
+CONVERSATION_SUMMARY_MODEL_VERSION = "latest";
 
 # How long to wait while polling batch transcription and conversation analysis status.
 WAIT_SECONDS = 10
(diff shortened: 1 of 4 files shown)


ut-karsh:
@trrwilson and @BrianMouncer for review.

scenarios/python/console/call-center/call_center.py
```diff
@@ -18,14 +18,14 @@
...
+SPEECH_TRANSCRIPTION_PATH = "/speechtotext/v3.1/transcriptions"
 
 # These should not change unless you switch to a new version of the Cognitive Language REST API.
 SENTIMENT_ANALYSIS_PATH = "/language/:analyze-text";
 SENTIMENT_ANALYSIS_QUERY = "?api-version=2022-05-01";
 CONVERSATION_ANALYSIS_PATH = "/language/analyze-conversations/jobs";
-CONVERSATION_ANALYSIS_QUERY = "?api-version=2022-05-15-preview";
-CONVERSATION_SUMMARY_MODEL_VERSION = "2022-05-15-preview";
+CONVERSATION_ANALYSIS_QUERY = "?api-version=2022-10-01-preview";
+CONVERSATION_SUMMARY_MODEL_VERSION = "latest";
```
BrianMouncer, line 28: what is the point of using latest, when CONVERSATION_ANALYSIS_QUERY and SPEECH_TRANSCRIPTION_PATH are static.

e.g. when v3.2 comes out, these existing clients will start using the new model version, but the old path and query, is that what you want to happen, or will that just make this more likely to break in the future?

BrianMouncer (approved):

scenarios/python/console/call-center/call_center.py
ut-karsh: This change was provided by the Language team. It is likely that the Conversation Summary model version is updated frequently even when the API path remains the same, which would be hard to keep track of.

PR #1830, opened 2023-02-10
Update embedded speech samples

pankopon:
## Purpose
Address a couple of issues in embedded speech samples:
* Speech SDK supports both full and short embedded model names, but the utility function in samples that checks the input parameters was not updated to take this into account, now fixed.
* When the C++ samples are run from Visual Studio IDE, the working directory (which is a user setting) needs to be configured so that the example data files are found. Added a note in README.

## Pull Request Type

```
[x] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

### DIFF
--- a/samples/cpp/embedded-speech/samples/settings.cpp
+++ b/samples/cpp/embedded-speech/samples/settings.cpp
@@ -238,7 +238,7 @@ bool VerifySettings()
                     }
                     else
                     {
-                        return model->Name.compare(modelName) == 0;
+                        return model->Name.compare(modelName) == 0 || model->Locales[0].compare(modelName) == 0;
                     }
                 });
 
@@ -298,7 +298,7 @@ bool VerifySettings()
                         }
                         else
                         {
-                            return voice->Name.compare(voiceName) == 0;
+                            return voice->Name.compare(voiceName) == 0 || voice->ShortName.compare(voiceName) == 0;
                         }
                     });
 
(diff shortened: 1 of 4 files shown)


yulin-li (approved):

samples/cpp/embedded-speech/samples/settings.cpp
```diff
@@ -238,7 +238,7 @@ bool VerifySettings()
                     }
                     else
                     {
-                        return model->Name.compare(modelName) == 0;
+                        return model->Name.compare(modelName) == 0 || model->Locales[0].compare(modelName) == 0;
```
BrianMouncer, line 241: >[0] [](http://example.com/codeflow?start=84&length=3)

Seems risky to assume a hard coded index.
Also, what happens if locales is empty?

BrianMouncer, line 241: >Locales [](http://example.com/codeflow?start=77&length=7)

If this is like windows sr and tts locales, it is possible to specify "en-US", "en-GB", or just "en" which would match any regional sub codes like "-US" or "-GB".

If that true here, and do you need to handle this?

BrianMouncer (approved):
:shipit:

samples/cpp/embedded-speech/samples/settings.cpp
pankopon: Model information comes from `GetSpeechRecognitionModels()`. SDK requires that a model has at least one locale (non-empty string), otherwise it is ignored when scanning for models. So model->Locales[0] is always there.

pankopon: Speech recognition locales are full-length, like "en-US". See https://learn.microsoft.com/azure/cognitive-services/speech-service/language-support?tabs=stt

The short form like "en" is for translation target languages:
https://learn.microsoft.com/azure/cognitive-services/speech-service/language-support?tabs=speech-translation

PR #2090, opened 2023-09-27
Fix datetime format

brandom-msft:
Minor formatting update (and also updating to the correct year 😁)

### DIFF
--- a/.github/workflows/label-stale-issues.yml
+++ b/.github/workflows/label-stale-issues.yml
@@ -2,7 +2,7 @@ name: label-stale-issues
 run-name: Issue freshness check
 on:
   schedule:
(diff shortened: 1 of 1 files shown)


.github/workflows/label-stale-issues.yml
```diff
@@ -2,7 +2,7 @@ name: label-stale-issues
 run-name: Issue freshness check
 on:
   schedule:
-    - cron: '0 2 * * 1-5'
+    - cron: '0/20 * * * 1-5'
```
brandom-msft, line 5: temporarily running every 20 minutes while debugging

glecaros (approved):

PR #2417, opened 2024-06-12
Make timer-based execution in IngestionClient a parameter rather than a variable.

superhindupur:
## Purpose
This PR makes the "Timer Based Execution" a deployment parameter rather than a variable so that users can change its value from the template deployment UI on Azure Portal rather than having to edit it in the ARM json template. 

I also updated the setup guide to make this choice clear.

I tested this by setting the Timer Based Execution parameter to both true and false and verifying the successful deployment in each case. 

When set to true, it deploys the StartTranscriptionByTimer function:
Screenshot 2024-06-12 153439

When set to false, it deploys the StartTranscriptionByServiceBus function:

Screenshot 2024-06-12 153443

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
[x] Other... Please describe: Deployment configuration and documentation. 
```


superhindupur:
Canceling this as we don't want users accidentally using non-timer based function that outs more load on the speech services and DB. If they want to use it on purpose they can always edit it from the ARM template.

PR #2423, opened 2024-06-14
Send completion notification to service bus queue provided by user in the Ingestion Client.

superhindupur:
## Purpose
This PR adds a feature to the IngestionClient to send a message to a queue provided by the user after successful completion of transcription and analytics. 

This message looks like the following, and consists of the file URIs for the original audio file and the output json file containing all the results from the IngestionClient. This notification is sent on the completion of a batch. 

```
[{"AudioFileName":"https://bhingclientstorage.blob.core.windows.net/audio-input/small_audio.wav","JsonReportLocation":"https://bhingclientstorage.blob.core.windows.net/json-result-output/small_audio.wav.json"},{"AudioFileName":"https://bhingclientstorage.blob.core.windows.net/audio-input/audio.wav","JsonReportLocation":"https://bhingclientstorage.blob.core.windows.net/json-result-output/audio.wav.json"}]
```

The user deploying IngestionClient will need to optionally provide the connection string to the specific queue they want this message to be sent to (if they leave it empty, this notification will not be sent anywhere). This has been added as an optional parameter in the deployment json file. 

Our team has a use case for this - we want to create further custom functions that will apply some more analytics on top of the transcripts as soon as the transcripts are ready. Our custom function(s) will be triggered by this service bus we provide. But I think this can benefit further users of IngestionClient. 

**NOTE** - I am opening this PR for early feedback. It is missing unit tests as of now - and I see that there are no existing unit tests for the areas I touched, so it is a bit involved. I will work with @HenryvanderVegte to add unit tests to this PR as a next step, as part of this same PR. 

## Pull Request Type

```
[ ] Bugfix
[x] Feature
[ ] Other... Please describe:
```


samples/ingestion/ingestion-client/.devcontainer/devcontainer.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "C# (.NET)",
-    "image": "mcr.microsoft.com/devcontainers/dotnet:1-8.0-bookworm",
+    "image": "mcr.microsoft.com/devcontainers/dotnet:1-6.0",
```
superhindupur, line 3: I saw that all of our .csproj files are configured for .NET 6.0, so I changed the devcontainer to use the same.

samples/ingestion/ingestion-client/Tests/EndToEndTests.cs
```diff
@@ -81,7 +81,7 @@ public async Task AnalyzeConversationTestAsync()
             var region = testProperties["LanguageServiceRegion"].ToString();
             var subscriptionKey = testProperties["LanguageServiceSubscriptionKey"].ToString();
             var provider = new AnalyzeConversationsProvider("en-US", subscriptionKey, region, Logger.Object);
-            var body = File.ReadAllText(@"testFiles/summarizationInputSample.json");
+            var body = File.ReadAllText(@"TestFiles/summarizationInputSample.json");
```
superhindupur, line 84: Changed the path to respect case-sensitivity so that it works on linux-based environments.

samples/ingestion/ingestion-client/FetchTranscription/TranscriptionProcessor.cs
```diff
@@ -429,7 +439,13 @@ private async Task ProcessSucceededTranscriptionAsync(string transcriptionLocati
...
                 var archiveFileLocation = System.IO.Path.GetFileNameWithoutExtension(fileName);
 
-                await StorageConnectorInstance.WriteTextFileToBlobAsync(editedTranscriptionResultJson, FetchTranscriptionEnvironmentVariables.JsonResultOutputContainer, jsonFileName, log).ConfigureAwait(false);
+                var jsonFileUrl = await StorageConnectorInstance.WriteTextFileToBlobAsync(editedTranscriptionResultJson, FetchTranscriptionEnvironmentVariables.JsonResultOutputContainer, jsonFileName, log).ConfigureAwait(false);
+
+                if (!string.IsNullOrEmpty(FetchTranscriptionEnvironmentVariables.CompletedServiceBusConnectionString))
+                {
+                    var completedMessage = new CompletedMessage(audioFileInfo.FileUrl, jsonFileUrl);
+                    await ServiceBusUtilities.SendServiceBusMessageAsync(this.completedServiceBusSender, completedMessage.CreateMessageString(), log, GetMessageDelayTime(serviceBusMessage.PollingCounter)).ConfigureAwait(false);
+                }
```
superhindupur, line 448: One important thing to note here is that the notification is sent for each audio file that is transcribed, this is not per-batch. I'd love to hear your thoughts/concerns on this - would you rather send this notification per batch, and why?

superhindupur: @HenryvanderVegte and I discussed offline that per-batch notification is better since it will result in generating fewer messages on the queue, and it pushes the choice onto the consumer of the message whether to process per-batch or per-file.

HenryvanderVegte (approved):

PR #2454, opened 2024-06-27
feat: Time interval as a parameter while deploying for StartTranscriptionByTimer function

### LINKED ISSUE #2448: [IngestionClient] Make the interval on the StartTranscriptionByTimer function configurable.
**Is your feature request related to a problem? Please describe.**
Currently, the IngestionClient runs every 2 minutes on a timer to start the transcription for new audio files added to the storage container. It is not possible to change this interval without changing the code itself.

**Describe the solution you'd like**
We want to make this interval configurable so that users can change this value at deployment-time as a parameter rather than having to edit the code.

komalg1:
## Purpose

* Resolves #2448 

## Pull Request Type
* This PR adds timer schedule as a configurable parameter for the StartTranscriptionByTimer azure function while deploying Ingestion Client. It sets the default value to `0 */2 * * * *` (2 minutes)

```
[ ] Bugfix
[x] Feature
[ ] Other... Please describe:
```

### DIFF
--- a/samples/ingestion/ingestion-client/Setup/ArmTemplateBatch.json
+++ b/samples/ingestion/ingestion-client/Setup/ArmTemplateBatch.json
@@ -115,6 +115,15 @@
                 "description": "Enter the address of your private endpoint here (e.g. https://mycustomendpoint.cognitiveservices.azure.com/) if you are connecting with a private endpoint"
             }
         },
+        "StartTranscriptionFunctionTimeInterval":
+        {
+            "defaultValue": "0 */2 * * * *",
+            "type": "String",
+            "metadata":
+            {
+                "description": "The time interval for the timer trigger in the StartTranscription function. The default value is every 2 minutes."
+            }
+        },
         "ProfanityFilterMode":
         {
             "defaultValue": "None",
@@ -937,6 +946,7 @@
                 "MaxPollingDelayInMinutes": "[variables('MaxPollingDelayInMinutes')]",
                 "Locale": "[parameters('Locale')]",
                 "MessagesPerFunctionExecution": "[variables('MessagesPerFunctionExecution')]",
+                "StartTranscriptionFunctionTimeInterval": "[parameters('StartTranscriptionFunctionTimeInterval')]",
                 "ProfanityFilterMode": "[parameters('ProfanityFilterMode')]",
                 "PunctuationMode": "[parameters('PunctuationMode')]",
                 "RetryLimit": "[variables('RetryLimit')]",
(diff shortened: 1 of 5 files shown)


samples/ingestion/ingestion-client/Setup/ArmTemplateBatch.json
```diff
@@ -115,6 +115,15 @@
                 "description": "Enter the address of your private endpoint here (e.g. https://mycustomendpoint.cognitiveservices.azure.com/) if you are connecting with a private endpoint"
             }
         },
+        "StartTranscriptionFunctionTimeInterval":
+        {
+            "defaultValue": "0 */2 * * * *",
```
superhindupur, line 120: I think we'd like to only ask for a numeric value of the timer interval since the users can be non-technical. So basically just accept the interval and generate the cron string programmatically.

superhindupur: Based on @komalg1's research, Azure Functions don't support dynamic cron expressions in the trigger attribute anyway.

So the choice comes down to exposing either a cron or a TimeSpan directly to the users as a parameter. While cron expressions allow more flexibility, TimeSpan is easier for non-technical users to understand. Given that IngestionClient is promoted as a no-code solution, we need to be careful about this tradeoff between ease of use vs flexibilty (keeping in mind the only flexibility that TimeSpan doesn't seem to provide as compared to Cron is obscure cases like day-of-month, year etc which might not be relevant in the case of the StartTranscription function). 

@HenryvanderVegte - would you like to chime in?

komalg1: There is no code change required if we want to use TimeSpan. Only condition is that the function needs to be running on an App Service Plan if TimeSpan has to be used as TimeSpan does not work with Consumption Plan.
Let me know your thoughts.

superhindupur: Received a comment offline from Henry: "If TimeSpan is not supported by Consumption Plan, I think it's fine to keep CRON expressions. Still supporting Consumption is more important I believe." So let's keep the change as it is for now. Thanks Komal for bearing with me :)

superhindupur:
@komalg1 could you please update the version number on https://github.com/Azure-Samples/cognitive-services-speech-sdk/blob/master/samples/ingestion/ingestion-client/Setup/ArmTemplateBatch.json#L250 and the same on the realtime json file (since batch and realtime are always released together even if one of them has changed) to v2.0.15? That should be the only required change. Thanks!

HenryvanderVegte (approved):

Issue #2446, opened 2024-06-26
[IngestionClient] Introduce dependency injection and make code more unit-testable. 


superhindupur:
**Is your feature request related to a problem? Please describe.**
Currently, the service dependencies in the IngestionClient code are declared as static variables (possibly because we want to instantiate them only once per application instance). This makes it hard to mock these dependencies and unit test the behavior of the IngestionClient code. 

**Describe the solution you'd like**
We would like to use a dependency injection framework and stop using static variables for service dependencies, so that we can easily mock them and unit test the code.

komalg1:
Pending items:
Dependency Injection for `TranscriptionProcessor`, `TranscriptionAnalyticsOrchestrator`, `BatchClient`

pankopon:
@HenryvanderVegte There are several "[IngestionClient]" items open, are these actual customer issues or more general backlog work items? Is there a plan for them? Please update.

HenryvanderVegte:
As discussed in https://github.com/Azure-Samples/cognitive-services-speech-sdk/issues/2521, moving item to internal backlog. Not an actual customer issue, but general backlog work item.

PR #2470, opened 2024-07-03
refactor: Adding dependency injection for StorageConnector

komalg1:
## Purpose

* This PR is the first step towards #2446 
* Injects StorageConnector as a dependency to FetchTranscription & StartTranscription.
* Adds Startup.cs for the azure functions to be able to register the dependencies going forward.
* Refactors the corresponding calls to AudioFileInfo & TranscriptionProcessor.
* Tested executing all the functions.
* Will add corresponding tests soon & update this PR.

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
[ ] Other... Please describe:
```


samples/ingestion/ingestion-client/StartTranscriptionByServiceBus/Startup.cs
```diff
@@ -0,0 +1,57 @@
...
+            var storageConnectionString = StartTranscriptionEnvironmentVariables.AzureWebJobsStorage;
+            var blobServiceClient = new BlobServiceClient(storageConnectionString);
+            var storageCredential = new StorageSharedKeyCredential(
+                GetValueFromConnectionString("AccountName", storageConnectionString),
+                GetValueFromConnectionString("AccountKey", storageConnectionString));
+
+            builder.Services.AddSingleton(blobServiceClient);
+            builder.Services.AddSingleton(storageCredential);
+
+            builder.Services.AddTransient<IStorageConnector, StorageConnector>();
```
superhindupur, line 39: @komalg1 could you explain the significance of creating this as a transient instance as opposed to a singleton like blobServiceClient and storageCredential?

komalg1: Singleton service is created once & can be shared safely across multiple threads. Basically it prevents overhead of creating same objects. In this case for both BlobServiceClient & StorageSharedKeyCredential, we don't want to repeatedly create & dispose blob client involving recreating the connections (using credentials) every time.
Transient service is created each time it is requested. Mostly for services which hold context or state specific to the operation/request. So for StorageConnector, it might hold some context specific to the request. Basically it will instantiate a new instance every time & therefore prevent any clash in case of concurrent requests & any internal state or config will not be shared.
 
Also as a general case, Singleton classes are efficient for usage in case of connections & secret management because we don't want to recreate the same connection objects.

samples/ingestion/ingestion-client/Connector/StorageConnector.cs
```diff
@@ -10,104 +10,51 @@ namespace Connector
...
-        public static string GetFileNameFromUri(Uri fileUri)
+        public string GetFileNameFromUri(Uri fileUri)
         {
             if (fileUri == null)
             {
                 throw new ArgumentNullException(nameof(fileUri));
             }
 
-            var pathParts = HttpUtility.UrlDecode(fileUri.AbsolutePath).Split('/').ToList();
+            var pathParts = System.Web.HttpUtility.UrlDecode(fileUri.AbsolutePath).Split('/').ToList();
```
HenryvanderVegte, line 38: nit: why don't you just keep the 'using System.Web;' ?

samples/ingestion/ingestion-client/Connector/StorageConnector.cs
```diff
@@ -10,104 +10,51 @@ namespace Connector
...
-            this.accountKey = GetValueFromConnectionString("AccountKey", storageConnectionString);
-
-            this.blobServiceClient = new BlobServiceClient(storageConnectionString);
+            this.blobServiceClient = blobServiceClient ?? throw new ArgumentNullException(nameof(blobServiceClient));
+            this.storageCredential = storageCredential ?? throw new ArgumentNullException(nameof(storageCredential));
+            this.log = log ?? throw new ArgumentNullException(nameof(log));
         }
 
-        public static string GetFileNameFromUri(Uri fileUri)
+        public string GetFileNameFromUri(Uri fileUri)
```
HenryvanderVegte, line 31: does not access instance data - keep it static

samples/ingestion/ingestion-client/Connector/StorageConnector.cs
```diff
@@ -10,104 +10,51 @@ namespace Connector
...
-            }
-
-            var fileExtension = Path.GetExtension(fileName);
-
-            return fileName.Substring(0, fileName.Length - fileExtension.Length);
+            return string.Join('/', cleanedPathParts.Skip(1));
         }
 
-        public static string GetContainerNameFromUri(Uri fileUri)
+        public string GetContainerNameFromUri(Uri fileUri)
```
HenryvanderVegte, line 43: same here - keep static

samples/ingestion/ingestion-client/Connector/StorageConnector.cs
```diff
@@ -117,28 +64,32 @@ public async Task<byte[]> DownloadFileFromContainer(string containerName, string
...
-            var sasToken = sasBuilder.ToSasQueryParameters(new StorageSharedKeyCredential(this.accountName, this.accountKey)).ToString();
+            var sasToken = sasBuilder.ToSasQueryParameters(this.storageCredential).ToString();
             return $"{blobContainerClient.GetBlobClient(fileName).Uri}?{sasToken}";
         }
 
-        public async Task<string> WriteTextFileToBlobAsync(string content, string containerName, string fileName, ILogger log)
+        public async Task<string> WriteTextFileToBlobAsync(string content, string containerName, string fileName)
         {
-            log.LogInformation($"Writing file {fileName} to container {containerName}.");
+            this.log.LogInformation($"Writing file {fileName} to container {containerName}.");
```
HenryvanderVegte, line 92: thanks for the clean up!

samples/ingestion/ingestion-client/FetchTranscription/Startup.cs
```diff
@@ -28,6 +32,32 @@ public override void Configure(IFunctionsHostBuilder builder)
...
                   options => SqlServerDbContextOptionsExtensions.UseSqlServer(options, FetchTranscriptionEnvironmentVariables.DatabaseConnectionString));
             }
+
+            var storageConnectionString = Environment.GetEnvironmentVariable("AzureWebJobsStorage");
+            var blobServiceClient = new BlobServiceClient(storageConnectionString);
+            var storageCredential = new StorageSharedKeyCredential(
+                GetValueFromConnectionString("AccountName", storageConnectionString),
+                GetValueFromConnectionString("AccountKey", storageConnectionString));
+
+            builder.Services.AddSingleton(blobServiceClient);
```
HenryvanderVegte, line 42: the functionshostbuilder is shared across function invocations, right?

samples/ingestion/ingestion-client/StartTranscriptionByServiceBus/Startup.cs
```diff
@@ -0,0 +1,57 @@
...
+        }
+
+        private static string GetValueFromConnectionString(string key, string connectionString)
```
HenryvanderVegte, line 42: duplicated func?

samples/ingestion/ingestion-client/StartTranscriptionByTimer/Startup.cs
```diff
@@ -0,0 +1,55 @@
...
+                GetValueFromConnectionString("AccountName", storageConnectionString),
+                GetValueFromConnectionString("AccountKey", storageConnectionString));
+
+            builder.Services.AddSingleton(blobServiceClient);
+            builder.Services.AddSingleton(storageCredential);
+
+            builder.Services.AddTransient<IStorageConnector, StorageConnector>();
+        }
+
+        private static string GetValueFromConnectionString(string key, string connectionString)
```
HenryvanderVegte, line 40: duplicated again - pls move to an extension in Connector for example

samples/ingestion/ingestion-client/StartTranscriptionByTimer/Startup.cs
```diff
@@ -0,0 +1,55 @@
...
+            var storageConnectionString = StartTranscriptionEnvironmentVariables.AzureWebJobsStorage;
+            var blobServiceClient = new BlobServiceClient(storageConnectionString);
+            var storageCredential = new StorageSharedKeyCredential(
+                GetValueFromConnectionString("AccountName", storageConnectionString),
+                GetValueFromConnectionString("AccountKey", storageConnectionString));
+
+            builder.Services.AddSingleton(blobServiceClient);
+            builder.Services.AddSingleton(storageCredential);
+
+            builder.Services.AddTransient<IStorageConnector, StorageConnector>();
```
HenryvanderVegte, line 37: ILogger is already registered by default in the DI?

HenryvanderVegte:
PR looks good, thanks a lot! 
Please do not forget to increment the version in the ARM template.

HenryvanderVegte (approved):

samples/ingestion/ingestion-client/Connector/StorageConnector.cs
komalg1: Updated

samples/ingestion/ingestion-client/Connector/StorageConnector.cs
komalg1: This method is now being referenced from different files and is part of the interface

samples/ingestion/ingestion-client/StartTranscriptionByServiceBus/Startup.cs
komalg1: Moved it to Connector

samples/ingestion/ingestion-client/Connector/StorageConnector.cs
HenryvanderVegte: in theory it does not need to be part of the interface - and we could move it to another static class. but fine for me to keep as is

HenryvanderVegte (approved):
:shipit:

Issue #2457, opened 2024-06-27
[Ingestion Client] Move away from Access Keys in preference of RBAC


ross-p-smith:
**Is your feature request related to a problem? Please describe.**
The Functions communicate with ServiceBus and Storage - both of these are using Access Keys

**Describe the solution you'd like**
This solution should utilise Role Based Access Control and remove the need for keys

**Describe alternatives you've considered**

**Additional context**
This will require the deployment to need an Owner/User Access Administrator to deploy Role Permissions

ross-p-smith:
This would be implemented by using https://learn.microsoft.com/en-us/azure/azure-functions/functions-identity-based-connections-tutorial-2 running as an identity and removing the keys from AppSettings - this would make the application much more secure.

pankopon:
@HenryvanderVegte There are several "[IngestionClient]" items open, are these actual customer issues or more general backlog work items? Is there a plan for them? Please update.

HenryvanderVegte:
As discussed in https://github.com/Azure-Samples/cognitive-services-speech-sdk/issues/2521, moving item to internal backlog. Not an actual customer issue, but general backlog work item.

PR #2476, opened 2024-07-07
[IngestionClient] Refactor - add ServiceBusClient via Dependency Injection instead of static clients

HenryvanderVegte:
## Purpose

* Refactor - add ServiceBusClient via Dependency Injection instead of static clients

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
[ ] Other... Please describe:
```


mahilleb-msft (dismissed):
:shipit:

samples/ingestion/ingestion-client/FetchTranscription/Startup.cs
```diff
@@ -28,6 +30,20 @@ public override void Configure(IFunctionsHostBuilder builder)
                 builder.Services.AddDbContext<IngestionClientDbContext>(
                   options => SqlServerDbContextOptionsExtensions.UseSqlServer(options, FetchTranscriptionEnvironmentVariables.DatabaseConnectionString));
             }
+
+            builder.Services.AddAzureClients(clientBuilder =>
+            {
+                clientBuilder.AddServiceBusClient(FetchTranscriptionEnvironmentVariables.StartTranscriptionServiceBusConnectionString)
```
zhouwangzw, line 36: In future we probably should switch to use AzureDefaultCredential, instead of connection string.

zhouwangzw (dismissed):

samples/ingestion/ingestion-client/FetchTranscription/Startup.cs
zhouwangzw: No need to handle it now. You can go ahead and merge

HenryvanderVegte: yes you're right, there's already a gh issue for it:
https://github.com/Azure-Samples/cognitive-services-speech-sdk/issues/2457

Adding the services via DI will make it easier to switch to managed identity

mahilleb-msft (approved):
:shipit:

samples/ingestion/ingestion-client/FetchTranscription/Program.cs
```diff
@@ -44,6 +46,20 @@ public static void Main(string[] args)
                     s.AddSingleton(blobServiceClient);
                     s.AddSingleton(storageCredential);
                     s.AddTransient<IStorageConnector, StorageConnector>();
+
+                    s.AddAzureClients(clientBuilder =>
```
komalg1, line 50: Is there any reason for using AddAzureClients over AddSingleton?

samples/ingestion/ingestion-client/FetchTranscription/TranscriptionProcessor.cs
```diff
@@ -65,6 +49,22 @@ public TranscriptionProcessor(IServiceProvider serviceProvider, IStorageConnecto
             {
                 this.databaseContext = this.serviceProvider.GetRequiredService<IngestionClientDbContext>();
             }
+
+            var serviceBusClientFactory = serviceProvider.GetRequiredService<IAzureClientFactory<ServiceBusClient>>();
```
komalg1, line 53: Should we not inject `IAzureClientFactory<ServiceBusClient>` instead of calling `GetRequiredService` ? It can cause issues when we need to mock ServiceBusClient for tests
Let me know your thoughts

komalg1 (commented):
Looks good. Just a couple of questions

samples/ingestion/ingestion-client/FetchTranscription/TranscriptionProcessor.cs
HenryvanderVegte: makes sense - fixed!

samples/ingestion/ingestion-client/FetchTranscription/Program.cs
HenryvanderVegte: no, I think both should work and do the same thing:
https://stackoverflow.com/questions/77613026/is-using-addazureclients-and-a-clientfactory-for-azure-servicebus-di-preferred-o

komalg1 (approved):

zhouwangzw (approved):

PR #2487, opened 2024-07-10
[IngestionClient] Complete Azure Functions Isolated Worker migration

HenryvanderVegte:
## Purpose

* Complete Azure Functions Isolated Worker migration

## Pull Request Type

```
[x] Bugfix
[ ] Feature
[ ] Other... Please describe:
```


samples/ingestion/ingestion-client/StartTranscriptionByServiceBus/Program.cs
```diff
@@ -18,8 +17,7 @@ public static class Program
         /// Main entry point of the function app.
         /// </summary>
         /// <param name="args"></param>
-        /// <returns>A <see cref="Task"/> representing the result of the asynchronous operation.</returns>
-        public static async Task Main(string[] args)
+        public static void Main(string[] args)
```
ross-p-smith, line 20: Any reason why we are not going async now?

ross-p-smith (approved):

.github/workflows/ingestion_client.yaml
```diff
@@ -53,17 +53,17 @@ jobs:
         run: dotnet publish -c Release -f net8.0 --no-self-contained --output ./StartTranscriptionByTimer samples/ingestion/ingestion-client/StartTranscriptionByTimer/StartTranscriptionByTimer.csproj
 
       - name: Zip StartTranscriptionByTimer
-        run: cd StartTranscriptionByTimer && zip -r ../StartTranscriptionByTimer.zip *
+        run: cd StartTranscriptionByTimer && zip -r ../StartTranscriptionByTimer.zip * .azurefunctions
```
mahilleb-msft, line 56: This should work as well (also below):

```suggestion
        run: cd StartTranscriptionByTimer && zip -r ../StartTranscriptionByTimer.zip .
```

(The glob `*` is interpreted by the Shell below calling `zip`. Default is not to expand hidden files; e.g., for Bash, `man bash` -> Pathname Expansion)

HenryvanderVegte: Fixed!

samples/ingestion/ingestion-client/StartTranscriptionByServiceBus/Program.cs
HenryvanderVegte: no - I was just following the sample apps while trying to figure out why it does not work and moved it to sync while doing so.

I don't think that it's necessary, but kept it in my PR anyway..

HenryvanderVegte:
> Feels great that this is working so just interested in what was it that got it working as it looks like there are 3-4 fixes here
> 
> * adding the .azurefunctions into the zip
> * downgrading the ServiceBus library
> * moving away from async
> * net8 in the AppService site.config

There were a couple of issues, and this PR fixes multiple at the same time

1)
functions not being loaded when deploying on azure => .azurefunctions missing in release binaries

2)
functions are being loaded, but seeing startup issues like 'Could not load file or assembly' => downgrading pkg versions

3)
Azure Portal still showed azure function running net framework 4.8 for some reason => not sure what's going on, but setting "netFrameworkVersion": "v8.0" fixed it


I don't think adding 'System.Threading.ExecutionContext' or 'FrameworkReference' in csproj files made a difference, but since it was mentioned in the migration guide, I decided to add it anyway

mahilleb-msft (approved):

PR #2501, opened 2024-07-17
refactor: Injecting environment variables instead of using static env file

komalg1:
## Purpose

* This PR is a step towards #2446 
* Adds an AppConfig class which is injected when function loads & supplies the function specific environment values.
* Removes the static environment files

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
[ ] Other... Please describe:
```


samples/ingestion/ingestion-client/FetchTranscription/TranscriptionProcessor.cs
```diff
@@ -323,21 +329,21 @@ private async Task WriteErrorReportAsync(string errorString, string jobName)
...
+                this.appConfig.ErrorReportOutputContainer,
                 errorTxtname).ConfigureAwait(false);
         }
 
         private async Task ProcessSucceededTranscriptionAsync(string transcriptionLocation, string subscriptionKey, TranscriptionStartedMessage serviceBusMessage, string jobName, ILogger log)
         {
             log.LogInformation($"Got succeeded transcription for job {jobName}");
 
-            var transcriptionAnalyticsOrchestrator = new TranscriptionAnalyticsOrchestrator(serviceBusMessage.Locale, log);
+            var transcriptionAnalyticsOrchestrator = new TranscriptionAnalyticsOrchestrator(serviceBusMessage.Locale, log, Options.Create(this.appConfig));
```
HenryvanderVegte, line 340: why do you use Options.Create(this.appConfig) here instead of just passing the appConfig?

samples/ingestion/ingestion-client/StartTranscriptionByServiceBus/StartTranscriptionByServiceBus.cs
```diff
@@ -23,39 +19,19 @@ namespace StartTranscription
...
         /// </summary>
         /// <param name="logger">The StartTranscriptionByServiceBus Logger</param>
-        /// <param name="storageConnector">Storage connector dependency</param>
-        /// <param name="serviceBusClientFactory">Azure client factory for service bus clients</param>
+        /// <param name="transcriptionHelper"></param>
         public StartTranscriptionByServiceBus(
             ILogger<StartTranscriptionByServiceBus> logger,
-            IStorageConnector storageConnector,
-            IAzureClientFactory<ServiceBusClient> serviceBusClientFactory)
+            IStartTranscriptionHelper transcriptionHelper)
```
HenryvanderVegte, line 31: nice - looks much cleaner now!

samples/ingestion/ingestion-client/StartTranscriptionByTimer/Config/AppConfig.cs
```diff
@@ -0,0 +1,83 @@
...
+            set => this.retryLimit = Math.Clamp(value, 1, Constants.MaxRetryLimit);
+        }
+
+        public int InitialPollingDelayInMinutes
+        {
+            get => this.initialPollingDelayInMinutes;
+            set => this.initialPollingDelayInMinutes = Math.Clamp(value, 2, Constants.MaxInitialPollingDelayInMinutes);
+        }
+
+        public int MaxPollingDelayInMinutes { get; set; } = Constants.DefaultMaxPollingDelayInMinutes;
```
HenryvanderVegte, line 53: I believe you got a Math.Clamp here in the FetchTranscription AppConfig

HenryvanderVegte (approved):

samples/ingestion/ingestion-client/FetchTranscription/TranscriptionProcessor.cs
komalg1: `Options.Create(this.appConfig)` wraps the `appConfig` to `IOptions<AppConfig>` which is expected as a constructor parameter for `TranscriptionAnalyticsOrchestrator`.
Directly passing appConfig breaks the pattern of IOptions which is now being used for managing the configuration.
Also eventually we should also move `TranscriptionAnalyticsOrchestrator` to be used as a dependency. When that happens we won't need to pass the appConfig.

samples/ingestion/ingestion-client/StartTranscriptionByTimer/Config/AppConfig.cs
komalg1: I saw `ClampInt` method being used from the `EnvironmentVariablesExtensions` class which is behaving same as `Math.Clamp`. So I updated the code accordingly in all the places. I could see it for both FetchTranscription & StartTranscription. Let me know if you feel something is missing or needs to be removed.

HenryvanderVegte: no, what I meant is: 
in FetchTranscriptionConfig you got
```
        public int MaxPollingDelayInMinutes
        {
            get => this.maxPollingDelayInMinutes;
            set => this.maxPollingDelayInMinutes = Math.Clamp(value, 2, Constants.DefaultMaxPollingDelayInMinutes);
        }
```
and here you got:
```
public int MaxPollingDelayInMinutes { get; set; } = Constants.DefaultMaxPollingDelayInMinutes;
```

komalg1: Ah, nice catch, Thanks! I have updated it.

ross-p-smith (approved):
Awesome

Fantastic work - I might write some Unit tests in my spare time now! :-)

HenryvanderVegte (approved):

PR #2503, opened 2024-07-17
ci: Unit test coverage as part of the CI/CD pipeline

komalg1:
## Purpose

* This PR completes #2447 
* It adds a job to run unit tests and collect coverage
* Adds code coverage as a PR comment
* Uploads the summary in build job

## Pull Request Type
What kind of change does this Pull Request introduce?

```
[ ] Bugfix
[ ] Feature
[ ] Code style update (formatting, local variables)
[] Refactoring (no functional changes, no api changes)
[ ] Documentation content changes
[x] Other... Please describe: Changes to our CI configuration files and scripts
```

PR #2515, opened 2024-07-22
[IngestionClient] Unit test coverage as part of the CI/CD pipeline

ross-p-smith:
## Purpose
This PR is actually #2503 but with the comments fixed. Original work has been carried out by @komalg1 

This adds Code Coverage report into the project so we can baseline the tests

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
[x] Code style update (formatting, local variables)
[ ] Other... Please describe:
```


ross-p-smith:
@HenryvanderVegte this PR is using the GITHUB_TOKEN which will not work on a fork, as you cannot access that Token from a fork. I believe the theory is that this workflow when run under the repo context (not fork) will have access to the token and will be able to write the Test Coverage as a comment in the PR

.github/workflows/ingestion_client.yaml
```diff
@@ -31,8 +38,55 @@ jobs:
...
+
+      - name: Install coverlet
+        run: dotnet tool install --global coverlet.console
+
+      - name: Install reportgenerator
+        run: dotnet tool install --global dotnet-reportgenerator-globaltool
+
+      - name: Run Unit Tests and Collect Coverage
+        run: |
+          dotnet test ./samples/ingestion/ingestion-client/Tests --configuration Release /p:CollectCoverage=true /p:CoverletOutputFormat=cobertura
```
HenryvanderVegte, line 59: please add --filter "TestCategory=UnitTest" to dotnet test command (see https://learn.microsoft.com/en-us/dotnet/core/testing/selective-unit-tests?pivots=mstest#syntax)

samples/ingestion/ingestion-client/versions.nugets.props
```diff
@@ -6,6 +6,8 @@
 
     <MSTestTestAdapterPackageVersion>[3.4.3]</MSTestTestAdapterPackageVersion>
     <MSTestTestFrameworkPackageVersion>[3.4.3]</MSTestTestFrameworkPackageVersion>
+    <MSTestSdk>[17.10.0]</MSTestSdk>
```
HenryvanderVegte, line 9: nit of a nit: please use format of other packages (PackageVersion suffix)

samples/ingestion/ingestion-client/infra/main.json
```diff
@@ -4,11 +4,26 @@
...
   "parameters": {
+    "EnvironmentName": {
+      "type": "string",
+      "minLength": 1,
+      "maxLength": 20,
+      "metadata": {
+        "description": "Name of the the environment which is used to generate a short unique hash used in all resources."
+      }
+    },
+    "ResourceToken": {
```
HenryvanderVegte, line 20: what's the advantage of using ResourceToken here instead of DeploymentId with default utcNow()?

I think it's more likely that customers would accidentally create resources with the same ResourceToken if they pick an EnvironmentName themselves.. Maybe we should rename it keep it as DeploymentId with default utcNow(), and then generate the hash based on `[toLower(uniqueString(subscription().id, parameters('DeploymentId'), parameters('AzureSpeechServicesRegion')))]` if that makes sense to you?

.github/workflows/ingestion_client.yaml
ross-p-smith: Added

samples/ingestion/ingestion-client/infra/main.json
ross-p-smith: OK - this was actually an upcoming change that should probably be done in a separate PR so I have pulled this out so we can concentrate on Code Coverage which this PR is about. I have left the shortened Function Names in as that is a warning in App Service at the moment (but also happy to pull that out into a separate PR too)

HenryvanderVegte (approved):

zhouwangzw (approved):

yulin-li:
@ross-p-smith @HenryvanderVegte @zhouwangzw  The azure pipeline build from a forked repo is disabled now for some security reasons, we are contact to ado org admin to see if we can enable it again.

HenryvanderVegte:
> @ross-p-smith @HenryvanderVegte @zhouwangzw The azure pipeline build from a forked repo is disabled now for some security reasons, we are contact to ado org admin to see if we can enable it again.

thanks a lot @yulin-li !

PR #2605, opened 2024-09-25
[PA][Python]Update continuous mode with reference text when enable miscue

yaden-ms:
## Purpose

* To get miscue on omission and insertion for zh-cn like locales which need word segmentation.

## Pull Request Type

```
[x] Bugfix
[ ] Feature
[ ] Other... Please describe:
```


yaden-ms:
@microsoft-github-policy-service agree company="Microsoft"
________________________________

samples/python/console/speech_sample.py
```diff
@@ -37,6 +37,7 @@
 weatherfilenamemp3 = "whatstheweatherlike.mp3"
 weatherfilenamemulaw = "whatstheweatherlike-mulaw.wav"
 seasonsfilename = "pronunciation_assessment_fall.wav"
+zhcnfilename = "zhcn_sample.wav"
```
wangkenpu, line 40: As we synced before, this file should be a short audio sample.

1. Can you rename it as `zhcn_short_dummy_sample.wav`?
2. Can you check in this wave file?

samples/python/console/speech_sample.py
```diff
@@ -843,6 +844,44 @@ def pronunciation_assessment_from_microphone():
...
 
+def get_reference_words(wave_filename, reference_text, language):
+    audio_config = speechsdk.audio.AudioConfig(filename=wave_filename)
+    speech_config = speechsdk.SpeechConfig(subscription=speech_key, region=service_region)
+    speech_recognizer = speechsdk.SpeechRecognizer(
+        speech_config=speech_config, language=language, audio_config=audio_config
+    )
+
+    # Create pronunciation assessment config, set grading system, granularity and if enable miscue based on your requirement.
+    enable_miscue = True
```
wangkenpu, line 855: In order to get the result with word segmentation, `enable_miscue` should be set as `false`. Otherwise, some reference words will be ignored.

samples/python/console/speech_sample.py
```diff
@@ -843,6 +844,44 @@ def pronunciation_assessment_from_microphone():
...
+    pronunciation_config = speechsdk.PronunciationAssessmentConfig(
+        reference_text=reference_text,
+        grading_system=speechsdk.PronunciationAssessmentGradingSystem.HundredMark,
+        granularity=speechsdk.PronunciationAssessmentGranularity.Phoneme,
+        enable_miscue=enable_miscue)
+
+    # Apply pronunciation assessment config to speech recognizer
+    pronunciation_config.apply_to(speech_recognizer)
+    result = speech_recognizer.recognize_once_async().get()
```
wangkenpu, line 864: Here we should use continuous mode to get the reference word with word segmented results. Otherwise, we do not need to call the service twice.

samples/python/console/speech_sample.py
```diff
@@ -843,6 +844,44 @@ def pronunciation_assessment_from_microphone():
...
+        return None
+    elif result.reason == speechsdk.ResultReason.Canceled:
+        cancellation_details = result.cancellation_details
+        print("Speech Recognition canceled: {}".format(cancellation_details.reason))
+        if cancellation_details.reason == speechsdk.CancellationReason.Error:
+            print("Error details: {}".format(cancellation_details.error_details))
+        return None
+
+
 def pronunciation_assessment_continuous_from_file():
```
wangkenpu, line 885: Can we add a long zh-CN case for this sample?

samples/python/console/speech_sample.py
yaden-ms: The content in short wave files is different from that in reference text, we need to use error_type to filter out some insertion words?

samples/python/console/speech_sample.py
yaden-ms: 1) Using continuous mode can't get the word segmentation of reference text, as the final outputs will be recognized words.
2) The input fake wave is short, no need to use continuous mode.

samples/python/console/speech_sample.py
wangkenpu: Set `enable_miscue` as `False` will use linear grammar and to a force alignment, which means all reference words will be segmented. Here `error_type` is not useful.

samples/python/console/speech_sample.py
wangkenpu: make sense.

samples/python/console/speech_sample.py
yaden-ms: updated.

samples/python/console/speech_sample.py
yaden-ms: If set enable_miscue as False, PA engine will return speechsdk.ResultReason.NoMatch, as the content of the short wave file is totally different from what is in the reference text, unless we use the first few seconds wave data from the long wave file which matches the reference text.

samples/python/console/speech_sample.py
yaden-ms: added

samples/python/console/speech_sample.py
wangkenpu: Got it. Let's have a test with more samples and make sure that setting `enable_miscue` as `True` can get the word segmentation results.

samples/python/console/speech_sample.py
```diff
@@ -37,6 +37,8 @@
 weatherfilenamemp3 = "whatstheweatherlike.mp3"
 weatherfilenamemulaw = "whatstheweatherlike-mulaw.wav"
 seasonsfilename = "pronunciation_assessment_fall.wav"
+zhcnfilename = "zhcn_short_dummy_sample.wav"
+zhcnlongfilename = "zhcn_long_dummy_sample.wav"
```
wangkenpu, line 41: maybe `zhcnlongfilename = "zhcn_continuous_mode_sample.wav"` is better.

yaden-ms: updated.

wangkenpu (approved):

PR #2656, opened 2024-11-07
Update conversation analysis api version.

v-bingyang:
image

When running the call center sample project, an error 401 is reported when requesting the API. After updating the api-version to 2024-05-15-preview, it becomes normal.


scenarios/csharp/dotnetcore/call-center/call-center/Program.cs
```diff
@@ -94,7 +94,7 @@ public class Program
         private const string sentimentAnalysisPath = "language/:analyze-text";
         private const string sentimentAnalysisQuery = "api-version=2022-05-01";
         private const string conversationAnalysisPath = "/language/analyze-conversations/jobs";
-        private const string conversationAnalysisQuery = "api-version=2022-10-01-preview";
+        private const string conversationAnalysisQuery = "api-version=2024-05-15-preview";
```
rhurey, line 97: Is there a GA version of this?

PR #2763, opened 2025-03-03
Remove Node.js Code from scenarios Folder as it's no longer maintained

centwang:
Remove Node.js Code from scenarios Folder as it's no longer maintained.


nbcsm (commented):
Also need to update this one at the bottom: https://github.com/Azure-Samples/cognitive-services-speech-sdk/blob/master/samples/batch/README.md

nbcsm (approved):

PR #3000, opened 2026-03-20
update samples to use 2026-01-01

ms-feizhao:
## Purpose

* update samples to use 2026-01-01

## Pull Request Type

```
[ ] Bugfix
[x] Feature
[ ] Other... Please describe:
```

### DIFF
--- a/samples/custom-voice/.gitignore
+++ b/samples/custom-voice/.gitignore
@@ -1,0 +1,2 @@
+# Environment files containing secrets - do not commit
+.env
--- a/samples/custom-voice/.vscode/launch.json
+++ b/samples/custom-voice/.vscode/launch.json
@@ -1,0 +1,33 @@
+{
+    "version": "0.2.0",
+    "configurations": [
+        {
+            "name": "Python: Personal Voice Sample",
+            "type": "debugpy",
+            "request": "launch",
+            "program": "${workspaceFolder}/python/personal_voice_sample.py",
+            "cwd": "${workspaceFolder}/python",
+            "envFile": "${workspaceFolder}/.env",
+            "console": "integratedTerminal"
+        },
+        {
+            "name": "Python: Professional Voice Sample",
+            "type": "debugpy",
+            "request": "launch",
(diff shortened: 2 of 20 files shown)


samples/custom-voice/.vscode/launch.json
```diff
@@ -0,0 +1,33 @@
+{
+    "version": "0.2.0",
+    "configurations": [
+        {
+            "name": "Python: Personal Voice Sample",
```
ForrestGumb, line 5: Do we need this file?

samples/custom-voice/.gitignore
```diff
@@ -0,0 +1,2 @@
+# Environment files containing secrets - do not commit
```
ForrestGumb, line 1: Do we need this and file above?

samples/custom-voice/.vscode/launch.json
ms-feizhao: Yes — this file provides pre-configured VS Code debug launch profiles for all three samples (Python Personal Voice, Python Professional Voice, and C# Custom Voice). It also wires up the .env file via envFile so users don't need to hardcode secrets or manually configure the debugger. It makes the "clone and run" experience much smoother.

samples/custom-voice/.gitignore
ms-feizhao: Yes, both files work together:

- .env.sample is a template showing all available environment variables with comments. Users copy it to .env and fill in their secrets (speech key, blob storage key, etc.).

- .gitignore ensures the actual .env file (which contains real secrets) is never accidentally committed to the repo.

This is a common pattern to avoid hardcoding credentials in source code (the old samples had "YourSubscriptionKey" placeholders directly in the code). Both the Python and C# samples now read from environment variables, and the .env file is the easiest way for users to configure them locally.

ForrestGumb (dismissed):

ms-feizhao:
#sign-off

yulin-li (approved):

